PR #3, opened 2021-11-24
Extend the OpenAPI spec for DQT

felixclack:
As part of the Find My TRN work, we need a way to find a person's
details, including their TRN, from a set of search parameters.

The DQT API currently exposes an endpoint that can be used to retrieve a
teacher's details if you already know their TRN. This isn't suitable for
our use case as we are supporting people who don't know their TRN.

I opted to add a new endpoint, `/trn_details`, rather than co-opt
the root `/teachers` resource. I made an assumption that the top-level
resource URI might want to be reserved for listing teachers when
search is not involved.

Also, this endpoint could return records for people that aren't
teachers, so coupling it to a teacher concept seems incorrect.

When helping someone retrieve their TRN, we ask them for an email
address first and attempt to match on that.

There will be another endpoint that a calling service can use to attempt
a TRN match using other parameters. This will be added in a future
change.

An assumption about the matching process is that we can't guarantee that
a single TRN record will be matched using these details. We are of
the belief that a TRN might match multiple people or a person might have
multiple TRN values.

This assumption means we expect the response of the new endpoint to
return an array of trn match objects. It will be up to the calling service
to disambiguate these results.

### DIFF
--- a/openapi/dqt.yml
+++ b/openapi/dqt.yml
@@ -1,0 +1,556 @@
+openapi: 3.0.1
+info:
+  title: DQT CRM API
+  description: 'API for Database of Qualified Teachers (DQT), allowing for retrieval of data for teacher including qualifications.'
+  version: v1
+servers:
+  - url: https://test-api-customerengagement.platform.education.gov.uk/dqt-crm/v1
+paths:
+  '/teachers/{trn}':
+    get:
+      summary: Teacher
+      description: Get an individual teacher by their trn
+      operationId: teacher
+      parameters:
+        - name: trn
+          in: path
+          description: Teacher Reference Number (TRN) e.g. 1001000
+          required: true
+          schema:
+            type: string
+        - name: birthdate
+          in: query
+          description: 'Birthdate in the format: yyyy-mm-dd'
+          required: true
+          schema:
+            type: string
+        - name: nino
+          in: query
+          description: National Insurance number - used if trn does not match e.g. A1A123A12
+          schema:
+            type: string
+        - name: Authorization
+          in: header
+          description: Bearer token authentication.
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: A teacher
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Teacher'
+              example:
+                application/json:
+                  trn: '1234567'
+                  ni_number: 1234AB12A
+                  name: Test Test
+                  dob: '1990-01-20T00:00:00.0000000+00:00'
+                  active_alert: false
+                  state: 0
+                  state_name: Active
+                  qualified_teacher_status:
+                    name: 'Qualified Teacher: Under the EC Directive'
+                    qts_date: '2018-07-05T23:00:00.0000000+00:00'
+                    state: 0
+                    state_name: Active
+                  induction:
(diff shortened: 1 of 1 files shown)


dendle (commented):
Looks good to me

aje54 (changes requested):
Suggest we change it to return a "trn details" record(s) for a person as opposed to "Teacher". The reason is that at the moment (will change eventually) non teachers can also have trns and this is a dqt API (not teacher). Number of ways we could skin  - I have had a think about an alternative suggestion to discuss  / consider...
rather than "teacher" the endpoint is /trndetails this returns an object that is an array of matches, I have used the name structure of dqt e.g. first name, middle name etc. and an array of emails (as there are many per trn row) - worth chewing over..

openapi: 3.0.0
info:
  version: 1.0.0
  title: teacher-identity
servers:
  - url: 'http://localhost:8080'
paths:
  /trndetails:
    get:
      summary: 'trn detils endpoint: returns trn details given an email'
      parameters:
        - name: email
          in: query
          description: email of person
          required: true
          schema:
            type: string         
      responses:
        '200':
          description: 'Ok'
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/trndetails'
                
components:
  #-------------------------------
  # Reusable schemas (data models)
  #-------------------------------
  schemas:
    #-------------------------------------------------
    # TRN details: Top level response model
    #-------------------------------------------------
    trndetails:
        type: object
        properties:
          matchingrecords:
       
            $ref: '#/components/schemas/trnmatches'
          
    trnmatches:            # Can be referenced as '#/components/schemas/trnmatches' 
      
      $ref: '#/components/schemas/trnmatch'

    emailaddresses:        # Can be referenced as '#/components/schemas/emailaddresses'
           
            $ref: '#/components/schemas/emailaddress'

    emailaddress:             # Can be referenced as '#/components/schemas/emailaddress'
      type: string
      example: 'email1@abc.com, email2@efg.com'
#-------------------------------------------------
    # trn match properties: 
#-------------------------------------------------

    trnmatch:              # Can be referenced as '#/components/schemas/trnmatch'
      type: object
      properties:
        trn:
          type: string
          example: '99456789'     # can add a format check

        emailaddresses:
          $ref: '#/components/schemas/emailaddresses'

        firstname:
          type: string
          example: 'Jenny'
        
        middlename:
          type: string
          example: 'jo'

        lastname:
          type: string
          example: 'jones'

        dateofbirth:
          type: string
          format: date
          example: "1972-06-23"
        
        nino:
          type: string
          example: "NY456789C"
        
        ittestablishment:           # could this return multiple?
          type: string
          example: "A nice ITT provider"
              
    Error:            # Can be referenced as '#/components/schemas/Error'
      type: object
      properties:
        code:
          type: integer
        message:
          type: string

dendle (approved):
Email match endpoint looks good to me

aje54 (changes requested):
might be an idea to match the field names up for clarity, in dqt [firstname,middlename,lastname], add dfeta_previouslastname. Make establishments an an array of establishment (with ukprn, establishmentname). There is an argument to hold a cache of these (from another endpoint?)  - lets discuss with TRA. Also add in the "verbose"  parameters as per google sheet.

PR #4, opened 2021-11-24
added first draft of broker API spec

dendle:
# TeacherID Broker API - OpenAPI spec draft

As part of the TeacherID system, the TeacherID - Broker API is intended to be used by calling systems to query datasets in DfE to find related identifiers, so that they can use the data they already have about a user to discover other data in other services related to the same user.

There are 2 main endpoints:
1. Find
  This endpoint is used to discover other attributes related to a user, given the attributes that the caller already has.  For example, a service might query for a TRN related to an email address - or might query if we hold a NINO for a given TRN etc.
2. Associate
  This endpoint is used to allow services to share information they learn about their users with other services.  In this case, if a service learns of an association between an email address and a TRN for example, this endpoint is used to share that association, and make it available for other services.

### DIFF
--- a/docs/openapi/brokerapi.yml
+++ b/docs/openapi/brokerapi.yml
@@ -7,6 +7,8 @@ paths:
   /associate:
     post:
       summary: Associate data points regarding a user
+      security: 
+      - api_key: []
       description: >-
         Used by calling systems to share associations between user attributes
         and identifiers.  For example, connecting an email address to a TRN, or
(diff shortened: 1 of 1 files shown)


aje54 (changes requested):
please change to be OAS 3.0 spec. Using components for data objects https://swagger.io/docs/specification/components/

dendle:
@aje54 Spec version updated to 3.0.3

dendle:
Going to park the Broker API for now, in order to focus on FindMyTRN service

aje54 (approved):

PR #5, opened 2021-11-29
Stripped down C4 to remove broker API

dendle:
# Update to C4 Diagram to remove Broker API

This update removes the Broker API from the architecture, in place of a direct connection between the FindMyTRN app and the DQT API.  As a result, the GOV.UK Account interactions that were present are also removed.

More detail is also added to the FindMyTRN Web app.


aje54 (approved):
Reviewed no comments needed. Looks good.

PR #14, opened 2022-07-29
Authz server/Find integration

gunndabad:


docs/find-lost-trn-integration.md
```diff
@@ -0,0 +1,72 @@
+# Find a lost TRN integration with the authorization server
+
+When the `qualified_teacher` scope is requested by a client, the authorization server needs to integrate with the
```
gunndabad, line 3: @aje54 I've called the scope `qualified_teacher` in the spike as `dqt` seemed a little too implementation-specific.

docs/find-lost-trn-integration.md
```diff
@@ -0,0 +1,72 @@
...
+This redirect includes some additional context, specified as query parameters:
+
+| Query parameter | Remarks |
+| --- | --- |
+| email | The verified email address as captured by the authorization server. This allows Find a lost TRN to skip asking the user for their email address again. |
+| redirect_uri | This is the callback URL on the authorization server that Find should redirect the user to once it has resolved a TRN. |
+| client_title | The name of the client that initiated the authorization journey. This enables 'branding' both the the authorization server and Find a lost TRN such that the user perceives the journey as a single service e.g. 'Register for a National Professional Qualification'. |
+| sig | This is a signed hash of the previous query parameters using a pre-shared key. |
+
+### Context signature query parameter
```
gunndabad, line 41: An alternative to this would be to use a JWT, just like in the handover back from Find. I opted against that here since most of the context is not claims about the user so using a JWT felt a little smelly.

docs/find-lost-trn-integration.md
aje54: quailified_teacher scope should be where the record has QTS - is that what you mean? Otherwise we had spoken about a "trn_scope" which was the generic "find the record in DQT" scope that doesn't really care if they have QTS.

gunndabad: That makes sense. Sounds like `trn` is what we want here.

aje54: as per slack convo, use "trn" scope or similar

aje54 (approved):
change qts sope to trn scope and all good.

PR #36, opened 2022-08-12
Infra resources added for the service Get an Identity.

prabhakarreddy1234:
Create below resources with terraform and added required make rules.

postgres
redis
auth server
test client
dataprotectionkeys
app insights

[x] Rebase to master
[x] Attach to trello

### DIFF
--- a/terraform/app.tf
+++ b/terraform/app.tf
@@ -1,0 +1,138 @@
+resource "azurerm_service_plan" "service-plan" {
+  name                = "getanid-authserver-${var.environment_name}-app-spln"
+  location            = data.azurerm_resource_group.group.location
+  resource_group_name = data.azurerm_resource_group.group.name
+  os_type             = "Linux"
+  sku_name            = var.app_service_plan_sku
+
+  lifecycle {
+    ignore_changes = [
+      tags
+    ]
+  }
+}
+
+resource "azurerm_linux_web_app" "auth-server-app" {
+  name                = local.get_an_identity_app_name
+  location            = data.azurerm_resource_group.group.location
+  resource_group_name = data.azurerm_resource_group.group.name
+  service_plan_id     = azurerm_service_plan.service-plan.id
+  https_only          = true
+
+  identity {
+    type = "SystemAssigned"
+  }
+
+  site_config {
+    http2_enabled       = true
+    minimum_tls_version = "1.2"
+  }
+
+  app_settings = {
+    HOSTING_ENVIRONMENT                    = var.environment_name,
+    APPLICATION_INSIGHTS_CONNECTION_STRING = azurerm_application_insights.insights.connection_string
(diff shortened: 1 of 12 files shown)


prabhakarreddy1234:
[Create dev environment with terraform](https://trello.com/c/snrbKjdT/631-create-dev-environment-with-terraform)

terraform/app.tf
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  app_settings = {
+    HOSTING_ENVIRONMENT                    = var.environment_name,
+    APPLICATION_INSIGHTS_CONNECTION_STRING = azurerm_application_insights.insights.connection_string
+    REDIS_URL                              = "${azurerm_redis_cache.redis.hostname}/${azurerm_redis_cache.redis.primary_access_key}",
+    AZURE_STORAGE_ACCOUNT_NAME             = azurerm_storage_account.data-protection.name,
+    AZURE_STORAGE_ACCESS_KEY               = azurerm_storage_account.data-protection.primary_access_key,
+    AZURE_STORAGE_CONTAINER                = azurerm_storage_container.uploads.name
+  }
```
gunndabad, line 32: Can we get the Postgres connection string in here too?

terraform/outputs.tf
```diff
@@ -0,0 +1,3 @@
+output "get_an_identity_fqdn" {
+  value = "https://${azurerm_linux_web_app.auth-server-app.name}.azurewebsites.net"
```
gunndabad, line 2:
```suggestion
  value = "https://${azurerm_linux_web_app.auth-server-app.default_site_hostname}"
```

prabhakarreddy1234: `default_site_hostname` isn't exposed as an attribute. But `default_hostname` can be used. i will try and see what value it returns.

terraform/app.tf
prabhakarreddy1234: Any specific config key name that you want for  postgres connectionstring ?

gunndabad: > Any specific config key name that you want for postgres connectionstring ?

Ideally `ConnectionStrings__DefaultConnection`

gunndabad (approved):

merrilmathew (approved):

terraform/app.tf
```diff
@@ -0,0 +1,138 @@
...
+    type = "SystemAssigned"
+  }
+
+  site_config {
+    http2_enabled       = true
+    minimum_tls_version = "1.2"
+  }
+
+  app_settings = {
+    HOSTING_ENVIRONMENT                    = var.environment_name,
```
saliceti, line 32: Since this is the same map as the other app, can we create a local variable instead?

terraform/app.tf
```diff
@@ -0,0 +1,138 @@
...
+
+  lifecycle {
+    ignore_changes = [
+      tags
+    ]
+  }
+}
+
+resource "azurerm_linux_web_app" "test-server-app" {
+  count               = var.environment_name == "dev" ? 1 : 0
```
saliceti, line 49: Could we create a variable in de.tfvars.json instead and call it somthing like "deploy_test_server_app"? this would be true for dev and default to false

PR #40, opened 2022-08-15
Email the email confirmation PIN via Notify

gunndabad:
https://trello.com/c/WHPlaHaN/612-email-user-the-pin-via-notify

We also email the user from a Hangfire job instead of directly so we get retries.


gunndabad:
[Email user the PIN via Notify](https://trello.com/c/WHPlaHaN/612-email-user-the-pin-via-notify)

MrKevJoy (approved):

PR #47, opened 2022-08-17
Handle new users where a TRN is *not* matched in Find

gunndabad:
Find will not always locate the TRN so we need to handle the case where this is missing in the callback JWT.


gunndabad:
[Final Find lost TRN integration](https://trello.com/c/bvFuYzPa/613-final-find-lost-trn-integration)

MrKevJoy (approved):

PR #56, opened 2022-08-19
Finalize email verification behavior

gunndabad:
This change adjusts some of the email verification PIN behaviors based on the patterns at https://design-system.service.gov.uk/patterns/confirm-a-phone-number/ and https://design-system.service.gov.uk/patterns/confirm-an-email-address/

Specifically:
- PINs are 5 digits and last for 15 minutes.
- Generating a new PIN invalidates any previous PINs for the same email address.
- If a PIN is used that has expired but expired within the last two hours we generate and email a new PIN.
- Richer errors based on code too long, too short, not numeric etc.

This also moves the sending of the email into the `EmailVerificationService` instead of having the caller having to invoke `IEmailSender` separately.


gunndabad:
[Email confirmation with a PIN](https://trello.com/c/OLPKT1pm/611-email-confirmation-with-a-pin)

MrKevJoy (approved):

PR #60, opened 2022-08-19
Add postgres firewall rule

merrilmathew:
To allow communication between app service and postgres instance, a
firewall rule needs to be added to allow Azure service to reach the
postgres.

### DIFF
--- a/terraform/app.tf
+++ b/terraform/app.tf
@@ -123,6 +123,14 @@ resource "azurerm_postgresql_flexible_server_database" "postgres-database" {
   server_id = azurerm_postgresql_flexible_server.postgres-server.id
 }
 
+resource "azurerm_postgresql_firewall_rule" "postgres-firewall" {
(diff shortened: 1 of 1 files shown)


terraform/app.tf
```diff
@@ -123,6 +123,14 @@ resource "azurerm_postgresql_flexible_server_database" "postgres-database" {
   server_id = azurerm_postgresql_flexible_server.postgres-server.id
 }
 
+resource "azurerm_postgresql_firewall_rule" "postgres-firewall" {
+  name                = "AllowAzure"
+  resource_group_name = data.azurerm_resource_group.group.name
+  server_name         = local.postgres_server_name
```
prabhakarreddy1234, line 129: we can get the name from postgres server resource so implicit dependency gets established.  `azurerm_postgresql_flexible_server.postgres-server.name` . this will avoid chicken and egg issues when recreating the server.

PR #72, opened 2022-08-24
Include the client's display name in the email verification copy

gunndabad:


gunndabad:
[Email user the PIN via Notify](https://trello.com/c/WHPlaHaN/612-email-user-the-pin-via-notify)

MrKevJoy (approved):

PR #94, opened 2022-09-02
Postgres backup workflow added

prabhakarreddy1234:
Scheduled job that creates a backup of postgres every night and uploads it to azure storage.

[x] Rebase to main
[x] Attach to trello

### DIFF
--- a/.github/workflows/database-backup.yml
+++ b/.github/workflows/database-backup.yml
@@ -1,0 +1,102 @@
+name: Backup Database to Azure Storage
+
+on:
+  workflow_dispatch:
+  schedule: # 01:00 UTC
+    - cron: "0 1 * * *"
+
+jobs:
+  backup:
+    name: Backup Azure Database ( Production )
+    runs-on: ubuntu-latest
+    environment:
+      name: production
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - uses: Azure/login@v1
+        with:
+          creds: ${{ secrets.azure_credentials }}
+
+      - name: Set environment variables
+        shell: bash
+        run: |
+          tf_vars_file=terraform/workspace_variables/production.tfvars.json
+          echo "KEY_VAULT_NAME=$(jq -r '.key_vault_name' ${tf_vars_file})" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


prabhakarreddy1234:
[Postgres backup workflow](https://trello.com/c/q5GRPQBh/642-postgres-backup-workflow)

.github/workflows/database-backup.yml
```diff
@@ -0,0 +1,105 @@
...
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+
+      - name: Add firewall rule
+        uses: azure/CLI@v1
+        with:
+          azcliversion: 2.30.0
+          inlineScript: |
+            az postgres flexible-server firewall-rule create --name ${{ env.POSTGRES_SERVER_NAME }} --resource-group ${{ env.RESOURCE_GROUP_NAME }} --rule-name Allow-GithubRunner-Postgres --start-ip-address ${{ env.RUNNER_IP }} --end-ip-address ${{ env.RUNNER_IP }}
+
+      - name: Backup Production DB
```
saliceti, line 71: We could compress directly here to avoid the separate compres step. See: https://github.com/DFE-Digital/apply-for-teacher-training/blob/main/.github/workflows/database-backup.yml#L37
(@RMcVelia can help as he did all the testing)

saliceti:
Can we add slack notification in case of error in the workflow? I'm concerned if the workflow fails when the firewall is open, then it may stay open forever. The alert would tell us to remediate ASAP.

.github/workflows/database-backup.yml
```diff
@@ -0,0 +1,105 @@
...
+          --connection-string '${{ steps.get_storage_secret.outputs.BACKUP-STORAGE-CONNECTION-STRING }}' \
+          --overwrite true
+
+      - name: Notify Slack channel on job failure
+        if: failure()
+        uses: rtCamp/action-slack-notify@v2
+        env:
+          SLACK_USERNAME: CI Deployment
+          SLACK_TITLE: Database backup failure
+          SLACK_MESSAGE: Production database backup job failed
```
saliceti, line 102: Could we add something about checking the firewall rules?

saliceti (approved):
👏

PR #160, opened 2022-09-26
Move OAuth-related properties into their own type on AuthenticationState

gunndabad:
Also make them optional; we want to be able to sign in users outside of an OAuth authorization journey.


MrKevJoy (approved):

gunndabad:
[Add a simple mechanism for adding in staff users to Identity](https://trello.com/c/8QpneO8I/137-add-a-simple-mechanism-for-adding-in-staff-users-to-identity)

PR #193, opened 2022-10-05
Modify url test task

merrilmathew:
### Context
Website url for each environment should be checked on its custom domain to ensure that traffic is flowing to the recently deployed code. This enables to make sure that the most recent code is available on the code.


NickGraham101:
@merrilmathew I'll be fixing this in https://github.com/DFE-Digital/get-an-identity/pull/192

PR #201, opened 2022-10-11
Use a common User model in events

gunndabad:
For web hook notifications we want to be able to map from an event to a notification without a database lookup. To simplify this use a common `User` model in events and include enough data that we can generate a notification from it.


gunndabad:
[Notify clients when a user's details have changed](https://trello.com/c/v88Yc3x8/172-notify-clients-when-a-users-details-have-changed)

MrKevJoy (approved):

PR #226, opened 2022-10-18
Ping /status on slot swap

NickGraham101:
## Change

By default the App Service swap slot process pings the application root and accepts any status code as confirmation the staging app is ready to swap to production.  The app has implemented health checks so we should ping the `/status` endpoint for a 200 response which will provide more reliable confirmation the app is ready to be swapped to the production slot.

## Testing
Enabled blue green deployment for dev, temporarily upgraded app service plan to S1 and [deployed](https://github.com/DFE-Digital/get-an-identity/actions/runs/3274325523/jobs/5387770662) to that environment.


NickGraham101:
[Improve slot swap operations](https://trello.com/c/rgyVRBJ6/704-improve-slot-swap-operations)

merrilmathew (approved):

PR #268, opened 2022-11-04
Add option to use alternative Notify key for emails not in allow list

gunndabad:
For UR we have a bunch of dummy users in DQT, each with a dummy email. Since people can't access these dummy emails to retrieve email verification codes, we don't want to even attempt to send emails there. At the same time, we want people in the team to still get emails.

This change introduces an optional email domain allow list plus a fallback Notify client. When configured, if an email's domain is not on the allow list we will use the alternative Notify client, which will be configured to not send emails. For UR the Notify console can be used to retrieve the email verification codes.


gunndabad:
[Decide on method of retrieving pin verification emails for UR](https://trello.com/c/1X6vVgrK/294-decide-on-method-of-retrieving-pin-verification-emails-for-ur)

MrKevJoy (approved):

PR #333, opened 2022-12-21
Add permissions to build job

enformatique11:
### Context

The default permissions for GitHub Actions and dependabot triggered jobs are different, Actions have broader write permissions, dependabot has read-only. Explicitly setting the required permissions ensures that jobs triggered by either service will not encounter permissions errors.

Set permissions for the build job

### Guidance to review

It won't be possible to test whether this change fixes dependabot PRs until it is merged, though if the explicit permissions work for GitHub Actions they should also work for dependabot

### Checklist

-   [ ] Attach to Trello card

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,10 +3,10 @@ name: Build authorization server
 on:
   push:
     branches:
-    - main
+      - main
   pull_request:
     branches:
-    - main
+      - main
   workflow_dispatch:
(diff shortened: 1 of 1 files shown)


enformatique11:
[Fix dependabot builds](https://trello.com/c/G0FOLFY2/481-fix-dependabot-builds)

gunndabad (approved):

PR #338, opened 2022-12-22
Amend SetTeacherTrn endpoint to allow null

gunndabad:
`null` is when the support agent tried to find the TRN manually but could not. In this case, we set `TrnLookupStatus` to failed.


gunndabad:
[Add API support for when a user isn't found in DQT](https://trello.com/c/VvzoJLJo/479-add-api-support-for-when-a-user-isnt-found-in-dqt)

MrKevJoy (approved):

PR #354, opened 2023-01-11
Date of birth data fix

hortha:
### Context

The date of birth field in the identity users table previously defaulted to "-infinity" rather than NULL if no value was entered which has left the odd record in pre-prod with an invalid date of birth essentially and has caused an EF migration to fail.

- Data fix existing users records so that any date of birth values of "-infinity" become NULL.
- Amend default value of date of birth column to default to NULL. 

### Guidance to review

This is simply an Entity Framework change.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


hortha:
[Investigate Identity user denormalisation approaches for deduplication](https://trello.com/c/qcj49M7c/494-investigate-identity-user-denormalisation-approaches-for-deduplication)

gunndabad (approved):

PR #370, opened 2023-01-18
Add identity have NI number page for new TRN lookup flow

wilpow-github:
### Context

We're removing the dependency on Find from ID. This is the next page in the UI ported from Find.

Add Have NI number page

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/AuthenticationState.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/AuthenticationState.cs
@@ -84,6 +84,7 @@ public class AuthenticationState
     public string? TrnOwnerEmailAddress { get; private set; }
     [JsonInclude]
     public TrnLookupStatus? TrnLookupStatus { get; private set; }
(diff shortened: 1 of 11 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/AuthenticationState.cs
```diff
@@ -84,6 +84,7 @@ public class AuthenticationState
     public string? TrnOwnerEmailAddress { get; private set; }
     [JsonInclude]
     public TrnLookupStatus? TrnLookupStatus { get; private set; }
+    public bool HaveNationalInsuranceNumber { get; set; }
```
wilpow-github, line 87: I see everything else has a private setter, is there a reason we shouldn't make this public?

gunndabad: If we're setting directly rather than via a `SetXxx` method then that's fine.

gunndabad (approved):

PR #394, opened 2023-01-30
Fix API endpoints to work with client credentials grant

gunndabad:
### Context

The DQT API will need to talk to ID to associate DQT records with ID records. We need to use the `client_credentials` grant here since there's no admin user sign in involved in this process.

Amend API endpoints to allow the `CurrentUserId` to be `null` and add an `UpdatedByClientId` so we can at least track which client the change came through.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


MrKevJoy (approved):

gunndabad:
[Add client_credentials support to Identity](https://trello.com/c/8nS9kpAL/611-add-clientcredentials-support-to-identity)

PR #396, opened 2023-01-31
Add merge user support UI feature

wilpow-github:
### Context

As a TRA helpdesk agent
I want to merge two Identity user accounts
So that we can eliminate multiple user accounts for the same person

Update User model
Add new pages for merge user flow

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/ChooseTrn.cshtml
```diff
@@ -0,0 +1,26 @@
...
+}
+
+<div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds-from-desktop">
+        <form asp-page="ChooseTrn" method="post">
+            <govuk-radios asp-for="Trn">
+                <govuk-radios-fieldset>
+                    <govuk-radios-fieldset-legend class="govuk-fieldset__legend--xl"/>
+
+                    <govuk-radios-item value="@(Model.User!.Trn ?? "None")">@(Model.User!.Trn ?? "None")</govuk-radios-item>
```
wilpow-github, line 18: N/A is used on User page, but didn't seem to make sense here. Should we allow None as an option?

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/Confirm.cshtml
```diff
@@ -0,0 +1,41 @@
...
+                    <govuk-summary-list-row-value>@Model.User!.EmailAddress</govuk-summary-list-row-value>
+                </govuk-summary-list-row>
+                <govuk-summary-list-row>
+                    <govuk-summary-list-row-key>Name</govuk-summary-list-row-key>
+                <govuk-summary-list-row-value>@Model.User!.FirstName @Model.User!.LastName</govuk-summary-list-row-value>
+                </govuk-summary-list-row>
+                <govuk-summary-list-row>
+                    <govuk-summary-list-row-key>Trn</govuk-summary-list-row-key>
+                    <govuk-summary-list-row-value>@Model.GetMergeTrn()</govuk-summary-list-row-value>
+                    @if (Model.User!.Trn != Model.UserToMerge!.Trn)
```
wilpow-github, line 28: Only display the change action if the Trn's of the two users are different. In practice this is only the case when both are null

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/Confirm.cshtml.cs
```diff
@@ -0,0 +1,127 @@
...
+        if (!ModelState.IsValid)
+        {
+            return this.PageWithErrors();
+        }
+
+        MergeUsers(User!, UserToMerge!);
+
+        await _dbContext.SaveChangesAsync();
+
+        HttpContext.Session.Remove(MergeTrnKey);
```
wilpow-github, line 45: We want to unset this value to prevent unwanted interaction with subsequent merges

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/Confirm.cshtml.cs
```diff
@@ -0,0 +1,127 @@
...
+    private bool TryGetTrnForUpdate(User userToMerge, out string? trn)
+    {
+        trn = HttpContext.Session.GetString(MergeTrnKey);
+        if (trn is null)
+        {
+            return false;
+        }
+
+        trn = trn == "None" ? null : trn;
+        return trn == userToMerge.Trn;
```
wilpow-github, line 125: This is slightly awkward logic. Ideally want to return true if trn is different from current user.trn. However, also want to ensure that the trn we use is either user.Trn or userToMerge.Trn.

The statement as written covers both, even if it's not entirely clear

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/User.cshtml
```diff
@@ -56,6 +57,25 @@
...
+                <govuk-summary-list-row-value>
+                    @if (Model.MergedUserIds.IsNullOrEmpty())
+                    {
+                        <p>None</p>
+                    }
+                    else
+                    {
+                        foreach (var userId in Model.MergedUserIds!)
+                        {
+                            <p>@userId</p>
```
wilpow-github, line 71: This displays the full list, perhaps we want to be able to collapse/expand the list?

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/User.cshtml.cs
```diff
@@ -40,12 +40,16 @@ public UserModel(TeacherIdentityServerDbContext dbContext, IDqtApiClient dqtApiC
...
 
+    public IEnumerable<Guid>? MergedUserIds { get; set; }
+
     [FromRoute]
     public Guid UserId { get; set; }
 
     public async Task<IActionResult> OnGet()
     {
         var user = await _dbContext.Users
+            .IgnoreQueryFilters()
```
wilpow-github, line 51: This is fine here as we already expect this user not to be deleted

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/Confirm.cshtml.cs
```diff
@@ -0,0 +1,127 @@
...
+
+    private void MergeUsers(User user, User userToMerge)
+    {
+        userToMerge.MergedWithUserId = user.UserId;
+        userToMerge.IsDeleted = true;
+
+        var previouslyMergedUsers = userToMerge.MergedUsers ?? Enumerable.Empty<User>();
+        foreach (var previouslyMergedUser in previouslyMergedUsers)
+        {
+            previouslyMergedUser.MergedWithUserId = user.UserId;
```
wilpow-github, line 86: Update previously merged users with new user ID

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/ChooseTrn.cshtml
gunndabad: Agree, 'N/A' seems weird here - 'None' makes sense to me. Would be good to only have a single option here if neither user has a TRN, though (instead of getting two 'None' options).

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/ChooseTrn.cshtml.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using Microsoft.EntityFrameworkCore;
+using TeacherIdentity.AuthServer.Models;
+
+namespace TeacherIdentity.AuthServer.Pages.Admin.MergeUser;
+
+public class ChooseTrn : PageModel
+{
+    public new User? User { get; set; }
```
gunndabad, line 12: Nit: could we store just the properties we need here instead of the whole model objects?

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/ChooseTrn.cshtml.cs
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public IActionResult OnPost()
+    {
+        if (!ModelState.IsValid)
+        {
+            return this.PageWithErrors();
+        }
+
+        HttpContext.Session.SetString(Confirm.MergeTrnKey, Trn!);
```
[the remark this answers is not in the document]
gunndabad: Ok for now but would be good to ultimately get it into the URL (a query param, perhaps?)

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/SelectUser.cshtml.cs
```diff
@@ -0,0 +1,61 @@
...
+        var user = await _dbContext.Users.SingleOrDefaultAsync(u => u.UserId == UserId);
+
+        if (user is null || user.UserType != UserType.Default)
+        {
+            context.Result = NotFound();
+            return;
+        }
+
+        UserIds = _dbContext.Users
+            .Where(u => u.UserId != UserId && u.IsDeleted == false && u.UserType == UserType.Default)
```
gunndabad, line 56: Do we need the `IsDeleted` clause in here?

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/ChooseTrn.cshtml
wilpow-github: Ideally the user wouldn't see this page if both are None. You've seen my comment below! Is this ok as is?

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/ChooseTrn.cshtml.cs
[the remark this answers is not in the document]
wilpow-github: I figured that if we allow any other choices e.g. name/email then it would be odd to continue to pass around the chosen TRN in the query param, and that it ought to be stored somewhere server side

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/MergeUser/SelectUser.cshtml.cs
wilpow-github: No good spot, that was a hangover from changing the query filter

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/User.cshtml
gunndabad: Probably ok for now. Hopefully this list will be short!

gunndabad (approved):

gunndabad (approved):

PR #402, opened 2023-02-02
Normalize TRN and NINO before calling DQT API

gunndabad:
### Context

The Find endpoint on the DQT API requires TRNs and NINOs to be passed in exactly as they are stored in DQT. This adds normalization for both those two fields.

Normalize TRN and NINO before calling DQT API. Note we do this when performing the lookup rather than when the data is submitted so that the Check Answers page shows data exactly as the user entered it.

Also adds more tests around TRN lookup in general.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


gunndabad:

MrKevJoy (approved):

PR #403, opened 2023-02-02
Set HasPreviousNameOption via OnOfficialNameSet method

gunndabad:
### Context

We've converged on a pattern where each page sets the data on `AuthenticationState` via a single `OnXxxSet` method. The only place left where we had two calls that set state from a single page is on the `OfficialName` page.

Move the `HasPreviousNameOption` onto the `OnOfficialNameSet` method. Also moves the `HasPreviousNameOption` enum into `AuthenticationState`'s namespace.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


gunndabad:

MrKevJoy (approved):

PR #409, opened 2023-02-08
Tidy TRN Optional state checking

gunndabad:
Most pages in the 'TRN Optional' registration flow require previous pages to have been completed. Some also are prohibited in certain circumstances e.g. you cannot change your email after it's been verified. This PR attempts to make state checking simpler and more consistent.

This introduces an 'authentication milestone'. This is a point in a sign in/register journey when, once reached, the user cannot go back from. We use this as a simple check on most pages to ensure pages cannot be accessed out-of-order illegally (e.g. a user cannot enter their name before we have verified their email).

In addition, many pages check that the previous page has been completed and simply redirect to it if it has not. This covers cases like accessing the NI Number page before the DOB page has been completed.

There a few other minor changes for consistency too; we were using both `HaveXxx` and `HasXxx` in various places. This settles on `HasXxx`.

Also fixes the TRN Optional journey to handle cases where a TRN has been found but we already have a user allocated to it.


gunndabad:

MrKevJoy (approved):

PR #410, opened 2023-02-08
Add search input for Admin Users page

wilpow-github:
### Context

As a TRA helpdesk agent
I want to find users by their TRN, name or email
So that I can easily navigate to a user's details

Add search input for user properties

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/Users.cshtml.cs
```diff
@@ -117,4 +122,57 @@ public record UserInfo
...
+        {
+            filterPredicate.And(user => lookupStatus.Contains(user.TrnLookupStatus));
+        }
+
+        if (!string.IsNullOrEmpty(userSearch))
+        {
+            switch (userSearch)
+            {
+                case var searchString when new Regex(@"^\d{7}$").IsMatch(searchString):
+                    filterPredicate.And(user => user.Trn == searchString);
```
wilpow-github, line 140: Andrew mentioned these could also be searched for via UserSearchAttributes. Might be better performance wise, but not sure if it would read better. How concerned are we about performance?

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/Users.cshtml.cs
```diff
@@ -117,4 +122,57 @@ public record UserInfo
...
+        }
+        else
+        {
+            searchPredicate.Or(a => a.AttributeType == "last_name" && a.AttributeValue == names[0]);
+        }
+
+        return await _dbContext.UserSearchAttributes
+            .Where(searchPredicate)
+            .GroupBy(a => a.UserId)
+            .Where(g => g.Count() >= names.Length)
```
wilpow-github, line 174: If we have two names separated by a space, make sure both match

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/Users.cshtml.cs
```diff
@@ -117,4 +122,57 @@ public record UserInfo
...
+
+        return filterPredicate;
+    }
+
+    private async Task<Guid[]> GetUsersByName(string searchString)
+    {
+        var names = searchString.Split(' ').Take(2).ToArray();
+
+        var searchPredicate = PredicateBuilder.New<UserSearchAttribute>(
+            a => a.AttributeType == "first_name" && a.AttributeValue == names[0]);
```
wilpow-github, line 160: Searching for exact match with names. Can do a contains, but may need to consider ranking the matches

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/Users.cshtml.cs
gunndabad: There's a unique index on `trn` so I suspect this is easily fast enough.

gunndabad (approved):

PR #411, opened 2023-02-08
Ensure email prefix valid for TRN Sign-In journey

wilpow-github:
### Context

As an Identity administrator
I want to prevent users signing up with shared email addresses
So that users do not lose access to their account when they change jobs

Validate supplied Email does not have invalid prefix in list

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Email.cshtml.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Email.cshtml.cs
@@ -1,7 +1,7 @@
 using System.ComponentModel.DataAnnotations;
-using Microsoft.EntityFrameworkCore;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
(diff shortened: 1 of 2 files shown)


gunndabad (approved):

gunndabad (changes requested):
Need to check if we already have a user with the email address before we fail; we don't want to stop any existing users from signing in, even if their email is shared (we can fix that later).

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Email.cshtml.cs
```diff
@@ -36,6 +41,16 @@ public async Task<IActionResult> OnPost()
             return this.PageWithErrors();
         }
 
+        if (_invalidEmailPrefixes.Contains(Email!.Split("@")[0]))
+        {
+            var existingUser = await _dbContext.Users.Where(user => user.EmailAddress == Email).SingleOrDefaultAsync();
```
wilpow-github, line 46: Not checking for Default User type as I believe this applies to all Users

gunndabad (approved):

PR #419, opened 2023-02-10
Reduce the width of the filters panel

gunndabad:
With production data the filter on the user list page is so wide some of the columns of the user list are truncated. This change reduces the width of the filter panel and reduces the size of the labels and checkboxes used within it.

It does mean our filter component CSS is non-standard but as we only use this in one place right now I think that's ok.


wilpow-github:
What does GDS say about forcing a table data to a single line via `white-space: nowrap;`? I'd only be worried about people with really long names

wilpow-github (approved):

gunndabad:
> What does GDS say about forcing a table data to a single line via `white-space: nowrap;`? I'd only be worried about people with really long names

I suspect it might be a problem on mobile, though I doubt there are many people accessing this admin page from their phones.

PR #430, opened 2023-02-16
Fix TrnLookupStatus for users not found in Find

gunndabad:
When we introduced `TrnLookupStatus` we set all existing users without a TRN to `Pending`, since NPQ expects all users to have a TRN.

For 'TRN Optional', the Find integration API endpoint was updated to add an additional property - `SupportTicketCreated` - so that for new users without a found TRN we would use `Pending` if a Zendesk support ticket were raised (since we expect the user to have a TRN based on their answers) or `None`, if the answers suggest they don't in fact have a TRN. Find never actually populated this property on the API however so the property was the default - `false` -  meaning we've marked all new users without a TRN `TrnLookupStatus.None`.

This change fixes up all users with `TrnLookupStatus.None` to make them `TrnLookupStatus.Pending` (I've checked and this is safe to do with the current production data). It also modifies the behaviour for new users to make them all `TrnLookupStatus.Pending`. 

When 'TRN Optional' rolls out proper we'll be using our own journey, not integrating with Find, so we will correctly derive `TrnLookupStatus` from then on.

I've also added a check constraint to ensure we're setting `TrnLookupStatus` when we need to.

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Migrations/20230216161649_TrnLookupStatusCheckConstraint.Designer.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Migrations/20230216161649_TrnLookupStatusCheckConstraint.Designer.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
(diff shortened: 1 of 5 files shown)


wilpow-github:
Presumable find a lost trn is raising support tickets. Do we not care about the resolution of those tickets for now?

gunndabad:
> Presumably find a lost trn is raising support tickets. Do we not care about the resolution of those tickets for now?

Tickets are not raised currently.

wilpow-github (approved):

PR #434, opened 2023-02-20
Check for TrnHolder in UserRequirements for pages in the TRN journey

gunndabad:
For the foreseeable future we will have distinct journeys for clients that need and TRN and those that do not. (This is identified by the presence of `TrnHolder` in `UserRequirements` on `AuthenticationState`). We need to make sure that users cannot access pages in the 'wrong' journey.

This change adds a filter to all pages under the `/SignIn/Trn/` folder that checks for `TrnHolder` and returns a 403 if it's missing.

Something similar can be done later for the `/SignIn/Register` journey but it's not included here.

### DIFF
--- a/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/EndpointTests/SignIn/TestBase.CommonTests.cs
+++ b/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/EndpointTests/SignIn/TestBase.CommonTests.cs
@@ -167,12 +167,12 @@ public partial class TestBase
     }
 
     public async Task ValidRequest_RendersContent(
+        Func<Configure, Func<AuthenticationState, Task>> configureAuthenticationHelper,
         string url,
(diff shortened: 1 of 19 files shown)


dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/EndpointTests/SignIn/TestBase.CommonTests.cs
```diff
@@ -167,12 +167,12 @@ public async Task MissingAuthenticationState_ReturnsBadRequest(HttpMethod method
     }
 
     public async Task ValidRequest_RendersContent(
+        Func<Configure, Func<AuthenticationState, Task>> configureAuthenticationHelper,
```
gunndabad, line 170: I flipped the parameter order here so this method is consistent with the others.

wilpow-github (approved):

PR #435, opened 2023-02-20
Add new previous user ID claim to access and identity tokens

wilpow-github:
### Context

We now support merging user accounts. When that happens one account is deleted and the other is retained (the ‘master’) with a link to the deleted record.

We send out a web hook with the ID of the deleted account and its new master so that consuming services can update their records and replace the old ID with the new. However, there’s a chance that services may not have processed the web hook by the time the user signs in again. If that happens, the consuming service may see a new ID and will incorrectly create a new user record.

Add an additional claim - previous_user_id - to the ID and access tokens that contains the IDs of merged accounts, if any

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Controllers/AuthorizationController.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Controllers/AuthorizationController.cs
@@ -380,6 +380,10 @@ public class AuthorizationController : Controller
 
                 yield break;
 
+            case CustomClaims.PreviousUserId:
+                yield return Destinations.IdentityToken;
+                yield break;
+
             default:
                 yield break;
         }
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/UserClaimHelper.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/UserClaimHelper.cs
@@ -1,4 +1,5 @@
 using System.Security.Claims;
+using LinqKit;
 using Microsoft.EntityFrameworkCore;
 using TeacherIdentity.AuthServer.Models;
 using TeacherIdentity.AuthServer.Oidc;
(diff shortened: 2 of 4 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/Controllers/AuthorizationController.cs
```diff
@@ -380,6 +380,10 @@ private static IEnumerable<string> GetDestinations(Claim claim, ClaimsPrincipal
 
                 yield break;
 
+            case CustomClaims.PreviousUserId:
```
wilpow-github, line 383: Add the claim no matter the scope?

dotnet-authserver/src/TeacherIdentity.AuthServer/UserClaimHelper.cs
```diff
@@ -82,6 +82,13 @@ public async Task<IReadOnlyCollection<Claim>> GetPublicClaims(Guid userId, Func<
         }
 #pragma warning restore CS0618 // Type or member is obsolete
 
+        var mergedUserIds = await _dbContext.Users.IgnoreQueryFilters()
+            .Where(u => u.MergedWithUserId == userId)
+            .Select(u => u.UserId)
+            .ToListAsync();
+
+        claims.Add(new Claim(CustomClaims.PreviousUserId, String.Join(",", mergedUserIds)));
```
wilpow-github, line 90: Should we not add the claim at all if there are no merged users?

gunndabad: I believe you can add multiple claims of the same type so perhaps one per ID (so zero if no merged users)

wilpow-github: Looks like if you add multiple claims of the same type they are combined into an array

gunndabad (approved):

PR #437, opened 2023-02-21
Fix 'TRN in use' email selection for new TRN journey

gunndabad:
We are missing an e2e test that covers 'TRN in use' for the new TRN journey. Adding it showed a bug; we were assuming we have a `JourneyTrnLookupState` around but we don't in this new journey. This change fixes that bug and adds an e2e test.

Once the new TRN journey is out of the door we can remove all the `JourneyTrnLookupState`-related bits and simplify things significantly.

### DIFF
--- a/dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/SignIn.StaffUser.cs
+++ b/dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/SignIn.StaffUser.cs
@@ -1,0 +1,113 @@
+using Microsoft.Playwright;
+using TeacherIdentity.AuthServer.Models;
+using TeacherIdentity.AuthServer.Oidc;
+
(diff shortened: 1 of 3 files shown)


dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/SignIn.StaffUser.cs
```diff
@@ -0,0 +1,113 @@
+using Microsoft.Playwright;
+using TeacherIdentity.AuthServer.Models;
+using TeacherIdentity.AuthServer.Oidc;
+
+namespace TeacherIdentity.AuthServer.EndToEndTests;
+
+public partial class SignIn
```
gunndabad, line 7: I've moved the staff sign in-related tests into their own file.

wilpow-github (approved):

PR #456, opened 2023-03-07
Add send confirm existing account phone page for register journey

wilpow-github:
### Context

As an overseas qualified teacher
I want to sign in/register a Teaching services account
So that I can apply for QTS

Add a new page at /sign-in/register/send-confirm-existing-account-phone

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Services/Redactor.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Services/Redactor.cs
@@ -63,6 +63,11 @@ public class Redactor
             return string.Empty;
         }
 
+        if (mobileNumber.Length <= 4)
+        {
+            return "***" + mobileNumber[^1];
+        }
+
         return new string('*', mobileNumber.Length - 4) + mobileNumber.Substring(mobileNumber.Length - 4);
     }
 }
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Services/Redactor.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Services/Redactor.cs
@@ -55,4 +55,14 @@ public class Redactor
 
         return builder.ToString();
     }
+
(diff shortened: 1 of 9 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/Services/Redactor.cs
```diff
@@ -55,4 +55,14 @@ public class Redactor
...
     }
+
+    public string? RedactMobileNumber(string mobileNumber)
+    {
+        if (string.IsNullOrEmpty(mobileNumber))
+        {
+            return string.Empty;
+        }
+
+        return new string('*', mobileNumber.Length - 4) + mobileNumber.Substring(mobileNumber.Length - 4);
```
gunndabad, line 66: Would be good to clone the implementation we're already using elsewhere: https://github.com/DFE-Digital/dqt-web-portal/blob/3811d7449b34f318d27f9e250f76a9181e890f1d/src/Dqt.Portals/Capgemini.DfE.Evolved.Model/RedactionHelper.cs#L71

wilpow-github: Would that work for international numbers? We're storing in the + format so the user will see e.g. +9************73. Could potentially cause confusion - especially if they entered as 0093...

gunndabad: Fair point. Let's keep it as you have it.

gunndabad (approved):

dotnet-authserver/src/TeacherIdentity.AuthServer/Services/Redactor.cs
wilpow-github: Added a case for when the mobile number is 4 or less

PR #461, opened 2023-03-10
Make event publishing async

gunndabad:
### Context

Currently we publish events from our `DbContext` when `SaveChanges{Async}` is called. In practice, publishing an event may involve sending messages to Service Bus. It’s possible that this Service Bus message publishing fails. If that happens, we would get an exception from `SaveChanges()` which would bubble up to the user and would suggest that their change has failed. In reality the data would have been committed to the database (but we wouldn’t notify anybody since generating web hooks would fail).

Don't called `IEventObserver.OnEventSaved` directly from `SaveChanges()`. Instead use a `BackgroundService` that polls the `events` table for unpublished events then calls `IEventObserver.OnEventSaved` for each one. 

For tests, use an interceptor that does immediately call `IEventObserver.OnEventSaved` (so we can `Assert` events have been emitted in tests without timing issues).

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


gunndabad:
[Make event publishing async](https://trello.com/c/6L93bWKp/699-make-event-publishing-async)

MrKevJoy (approved):

PR #464, opened 2023-03-14
Add Account landing page

gunndabad:
Identity needs an Account page that services can link to that enables a user to change their details.

This change adds the initial Account landing page that shows the user's ID and DQT details. Pages to actually update details will follow separately.

Since we need to be able to go 'back' to the calling service, we support passing in `client_id` and `redirect_uri` parameters, much like the OAuth authorization endpoint. We can be fairly permissive around checking `redirect_uri`; if the domain of the parameter matches the domain name of a redirect URI we have registered for that client then we're happy.

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/TeacherIdentityApplicationManager.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/TeacherIdentityApplicationManager.cs
@@ -52,18 +52,16 @@ public partial class TeacherIdentityApplicationManager : OpenIddictApplicationMa
         {
             var authority = new Uri(uri).GetLeftPart(UriPartial.Authority);
 
+            if (authority.Equals(addressAuthority))
+            {
+                return true;
+            }
+
             if (WildcardPathSegmentPattern().IsMatch(authority))
             {
                 var pattern = $"^{Regex.Escape(authority).Replace("__", ".*")}$";
 
                 if (Regex.IsMatch(addressAuthority, pattern))
-                {
-                    return true;
-                }
-            }
-            else
-            {
-                if (authority.Equals(addressAuthority))
                 {
                     return true;
                 }
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/TeacherIdentityApplicationManager.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/TeacherIdentityApplicationManager.cs
@@ -36,6 +36,43 @@ public partial class TeacherIdentityApplicationManager : OpenIddictApplicationMa
         }
     }
 
+    public async ValueTask<bool> ValidateRedirectUriDomain(Application application, string address, CancellationToken cancellationToken = default)
+    {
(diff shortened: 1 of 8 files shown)


gunndabad:
[Account landing page](https://trello.com/c/ie9TqRaU/802-account-landing-page)

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Account/Index.cshtml
```diff
@@ -0,0 +1,84 @@
...
+            <govuk-summary-list-row>
+                <govuk-summary-list-row-key>Mobile number</govuk-summary-list-row-key>
+                <govuk-summary-list-row-value>@Model.MobileNumber</govuk-summary-list-row-value>
+                <govuk-summary-list-row-actions>
+                    <govuk-summary-list-row-action href="#" visually-hidden-text="mobile number">Change</govuk-summary-list-row-action>
+                </govuk-summary-list-row-actions>
+            </govuk-summary-list-row>
+        </govuk-summary-list>
+
+        @if (Model.SafeRedirectUri is not null && Model.ClientDisplayName is not null)
```
wilpow-github, line 77: Is there a reason the logic of this back button is different to the one at the top of the page?

dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/TeacherIdentityApplicationManager.cs
```diff
@@ -36,6 +36,43 @@ public override async ValueTask PopulateAsync(OpenIddictApplicationDescriptor de
...
+                var pattern = $"^{Regex.Escape(authority).Replace("__", ".*")}$";
+
+                if (Regex.IsMatch(addressAuthority, pattern))
+                {
+                    return true;
+                }
+            }
+            else
+            {
+                if (authority.Equals(addressAuthority))
```
wilpow-github, line 66: Would it be better to check for equality before trying to find a match with regex?

wilpow-github: Probably can drop the else statement as well

wilpow-github (approved):

dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/TeacherIdentityApplicationManager.cs
gunndabad: Initially it was for symmetry with the existing `ValidateRedirectUriAsync` method (which was written to be as close as the base implementation as possible). I've flipped to check the direct match first.

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Account/Index.cshtml
gunndabad: It was a hangover of making this page reachable from the `/Complete` page. Since we don't need that quite yet, I've aligned the conditions for both.

PR #481, opened 2023-03-23
Improve email address rendering on overflow

gunndabad:
In a few places we use a soft hyphen to improve the wrapping of long email addresses. This extends the usage to a few new pages added recently.

I've also revised the methods that retrieve summary list row values in tests so they ignore these soft hyphen characters.

Before: 
before

After:
after


wilpow-github:
Would this also be useful on the /admin/users page? Is there a reason we shouldn't be using it everywhere we display an email?

wilpow-github (approved):

gunndabad:
> Would this also be useful on the /admin/users page? Is there a reason we shouldn't be using it everywhere we display an email?

Good spot, added to admin views too.

PR #483, opened 2023-03-24
Simplify propagating client callback URLs for account pages

gunndabad:
The `/account` pages are linked to from a client. The client passes its `client_id` and a `redirect_uri` (and in future, a sign out URI). We use this to link back to the calling service from the account pages.

Up until this point we've used a `returnUrl` callback parameter on all pages to ensure we can always get back to the root `/account` page with all this context intact. However, this is becoming unwieldy, especially now we're looking at adding even more context (sign out URI). Some journeys would need to track multiple return URLs too; one for the success case and another for the cancel (or back button) case.

We have some prior art for dealing with this - the `asid` query parameter we use for passing around the ID for a user's `AuthenticationState`. This works fairly well with `IIdentityLinkGenerator` and in tests so I've opted for something similar here. In this case it's much simpler, since the state is not mutable and can easily fit into the URL.

This change introduces a `ClientRedirectInfo` class to store the client ID and its callback URLs. There's a middleware that looks for the corresponding query parameters, validates them and assigns a `ClientRedirectInfo` if they're valid. It also returns an error if they're not valid. From then on the state is passed around a single string, encrypted using an `IDataProtector`. This ensures it cannot be tampered with. The appropriate methods on `IIdentityLinkGenerator` have been amended to take an optional `ClientRedirectInfo`.

I initially implemented this as a custom model binder but there are situations where we'll want to access this data where we won't have access to model binding (e.g. the nav UI in our layout file).

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Infrastructure/Middleware/ClientRedirectInfoMiddleware.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Infrastructure/Middleware/ClientRedirectInfoMiddleware.cs
@@ -1,0 +1,78 @@
+using Microsoft.AspNetCore.DataProtection;
+using TeacherIdentity.AuthServer.Oidc;
+
+namespace TeacherIdentity.AuthServer.Infrastructure.Middleware;
+
+public class ClientRedirectInfoMiddleware
+{
+    private readonly IDataProtector _dataProtector;
+    private readonly ILogger<ClientRedirectInfoMiddleware> _logger;
+    private readonly RequestDelegate _next;
+
+    public ClientRedirectInfoMiddleware(
+        IDataProtectionProvider dataProtectionProvider,
+        ILogger<ClientRedirectInfoMiddleware> logger,
+        RequestDelegate next)
+    {
+        _dataProtector = dataProtectionProvider.CreateProtector(nameof(ClientRedirectInfo));
+        _logger = logger;
+        _next = next;
+    }
+
+    public async Task Invoke(HttpContext context, TeacherIdentityApplicationManager applicationManager)
+    {
+        ClientRedirectInfo? clientRedirectInfo = null;
+
+        // Look for an existing encoded ClientRedirectInfo query parameter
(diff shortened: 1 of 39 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/Infrastructure/Middleware/ClientRedirectInfoMiddleware.cs
```diff
@@ -0,0 +1,79 @@
...
+            if (!ClientRedirectInfo.TryDecode(encodedQueryParam!, _dataProtector, out clientRedirectInfo))
+            {
+                _logger.LogDebug("Failed decoding encoded query parameter: '{QueryParameter}'.", encodedQueryParam!);
+                OnInvalidRequest();
+                return;
+            }
+        }
+
+        // No encoded query parameter found, check for unencoded client_id and redirect_uri
+        if (context.Request.Query.TryGetValue("client_id", out var clientId) &&
```
wilpow-github, line 38: Possibly an elseif?

dotnet-authserver/src/TeacherIdentity.AuthServer/Infrastructure/Middleware/ClientRedirectInfoMiddleware.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                _logger.LogDebug("Invalid redirect URI '{RedirectUri}' specified for client: '{ClientId}'.", redirectUri!, clientId!);
+                OnInvalidRequest();
+                return;
+            }
+
+            clientRedirectInfo = new ClientRedirectInfo(_dataProtector, clientId!, redirectUri!);
+        }
+
+        if (clientRedirectInfo is not null)
```
wilpow-github, line 60: Can we just assert this?

gunndabad: I'd like to keep these pages usable without this redirect info for now.

PR #489, opened 2023-03-28
Handle invalid email suffix tests running concurrently

gunndabad:
We have a few tests around checking invalid email suffixes that seem to be racing on CI. This change removes the up-front check for the invalid domain in favour of catching the exception if the domain is already in place.


gunndabad:
@wilpow-github I think the `if` check might be where the race is; that's returning false then in between the new record being inserted another test adds it - super unlucky

wilpow-github:
Ah ok, in that case yeah catching the exception when it happens seems best!

PR #490, opened 2023-03-28
Add TRN requirement type enum to applications table

wilpow-github:
### Context

The new ‘core’ registration journey is shortly to support finding a TRN from a DQT record, much like the existing ‘TRN Optional’ journey that was modelled on Find a lost TRN. We need a way to distinguish clients that will use the new journey vs the old. In addition, we need a way to know whether a client requires a TRN or whether finding one is optional.

Add a TrnRequirementType enum with values Legacy, Required and Optional. Add a non-nullable property to Application and a DB column. The default value must be 0 (Legacy).

Update the AddClient and EditClient pages to support modifying this property via a Select component.

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master


dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/EditClient.cshtml
```diff
@@ -24,6 +23,14 @@
         </govuk-checkboxes-item>
     </govuk-checkboxes>
 
+    <govuk-radios asp-for="TrnRequired">
+        <govuk-radios-fieldset>
+            <govuk-radios-fieldset-legend class="govuk-fieldset__legend--m"/>
+            <govuk-radios-item value="@TrnRequirementType.Required">Yes, client requires a TRN</govuk-radios-item>
+            <govuk-radios-item value="@TrnRequirementType.Optional">No, client does not require a TRN</govuk-radios-item>
+        </govuk-radios-fieldset>
+    </govuk-radios>
```
wilpow-github, line 32: Using radios for edit-client and a checkbox for add-client. New clients have a yes/no choice for TRN required, so checkbox seemed appropriate. Existing apps have a legacy mode which we only want to overwrite if TRN required/optional explicitly selected

gunndabad: It would be good to show the 'legacy' option here just to be safe; trying to reduce the chances of someone accidentally opting in to this new, as yet unfinished, behaviour by accident.

gunndabad (approved):

PR #492, opened 2023-03-29
Add AuthenticationJourneyType

gunndabad:
We have several pages that are used in multiple journeys (e.g. `/sign-in/email` and `/sign-in/complete`) and these pages need to know which journey they're in (as the content is different). This change adds `AuthenticationJourneyType` and makes it gettable from `AuthenticationState`. The value is derived from `TrnRequirementType` and scope.

There were a few computed properties on `AuthenticationState` that were being serialized so I've `[JsonIgnore]`d them too.


dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Landing.cshtml
```diff
@@ -16,7 +16,7 @@
 
         <vc:client-scoped-partial view-name="_SignIn.Landing.{0}.Content" />
 
-        <govuk-button-link href="@LinkGenerator.Register()">Create an account</govuk-button-link>
```
gunndabad, line 19: The `Register` root page we had has gone from the designs.

wilpow-github (approved):

PR #493, opened 2023-03-29
Replace ProtectedString with a signature in the query string

gunndabad:
`ProtectedString` was built as a way of passing around data in the URL without it being tampered with. The problem is that a user could copy/paste a valid `ProtectedString` from one page and use in another and this would (incorrectly) succeed e.g. an `email` parameter could be used in place of a `firstName` parameter.

This PR introduces a mechanism for attaching a signature of the query parameters to URLs. The signature is calculated using a secret and when the next page loads it re-computes the signature and returns an error if what was computed does not match what was provided. This is the same approach we've used with Find a lost TRN integration so it can check the data we've passed has not been tampered with.

To use this, add a `[VerifyInSignature]` attribute to each property on the Razor Page model that should be checked. Corresponding `LinkGenerator` methods need to include the signature generation query parameter.

After this, our URL generation logic, especially in tests, is becoming a little messy. I've opted to leave tidying that up to a follow up PR.


dotnet-authserver/src/TeacherIdentity.AuthServer/Helpers/QueryStringSignatureHelper.cs
```diff
@@ -0,0 +1,49 @@
...
+
+    public bool VerifySignature(Url url, string[] parameterNames)
+    {
+        var sig = CalculateSignature(url, parameterNames);
+        return sig == url.QueryParams.SingleOrDefault(qp => qp.Name == SignatureParameterName).Value?.ToString();
+    }
+
+    private string CalculateSignature(Url url, string[] parameterNames)
+    {
+        var canonicalizedValues =
```
wilpow-github, line 32: Can we not just hash the whole thing?

dotnet-authserver/src/TeacherIdentity.AuthServer/Helpers/QueryStringSignatureHelper.cs
```diff
@@ -0,0 +1,49 @@
+using System.Security.Cryptography;
+using System.Text;
+using Flurl;
+
+namespace TeacherIdentity.AuthServer.Helpers;
+
+public class QueryStringSignatureHelper
+{
+    private const string SignatureParameterName = "sig";
```
wilpow-github, line 9: I think you also added an app setting for this key

wilpow-github: Ah, no that was for the secret key!

dotnet-authserver/src/TeacherIdentity.AuthServer/Helpers/QueryStringSignatureHelper.cs
gunndabad: We can! I thought it might cause problems in a few places but all the tests are still green so I'll go ahead with that.

wilpow-github (approved):

PR #497, opened 2023-04-04
Add nav to account pages

gunndabad:
This change adds the nav UI for the account pages. It has a link to the root account page and a sign out page. The sign out link is back to the client's `sign_out_uri`, where one is provided. Otherwise, it links to a local `/account/sign-out` page.

I've amended TestClient so it has an account link and sign out process that mirrors the type of thing we'll see for real services. 

I've also added e2e tests to cover signing out and accessing the root account page, both with and without this client context.

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/IdentityLinkGenerator.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/IdentityLinkGenerator.cs
@@ -15,188 +15,190 @@ public abstract class IdentityLinkGenerator
 
     protected QueryStringSignatureHelper QueryStringSignatureHelper { get; }
 
-    protected abstract string PageWithAuthenticationJourneyId(string pageName, bool authenticationJourneyRequired = true);
+    protected abstract string Page(string pageName, bool authenticationJourneyRequired = true);
 
-    public string CompleteAuthorization() => PageWithAuthenticationJourneyId("/SignIn/Complete");
+    public string CompleteAuthorization() => Page("/SignIn/Complete");
 
-    public string Reset() => PageWithAuthenticationJourneyId("/SignIn/Reset");
+    public string Reset() => Page("/SignIn/Reset");
 
-    public string Email() => PageWithAuthenticationJourneyId("/SignIn/Email");
+    public string Email() => Page("/SignIn/Email");
 
-    public string EmailConfirmation() => PageWithAuthenticationJourneyId("/SignIn/EmailConfirmation");
+    public string EmailConfirmation() => Page("/SignIn/EmailConfirmation");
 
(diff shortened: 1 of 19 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/IdentityLinkGenerator.cs
```diff
@@ -15,177 +15,180 @@ protected IdentityLinkGenerator(QueryStringSignatureHelper queryStringSignatureH
 
     protected QueryStringSignatureHelper QueryStringSignatureHelper { get; }
 
-    protected abstract string PageWithAuthenticationJourneyId(string pageName, bool authenticationJourneyRequired = true);
+    protected abstract string Page(string pageName, bool authenticationJourneyRequired = true);
```
gunndabad, line 18: I've renamed this since it's also used for regular page links (without the authentication journey ID appended).

dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/SignOut.cs
```diff
@@ -9,53 +9,96 @@ public class SignOut : IClassFixture<HostFixture>
...
+        await page.SignInFromLandingPage();
+
+        await page.SubmitEmailPage(user.EmailAddress);
+
+        await page.SubmitEmailConfirmationPage(_hostFixture.CapturedEmailConfirmationPins.Last().Pin);
+
+        await page.SignOutFromAccountPageWithoutClientContext();
+
+        _hostFixture.EventObserver.AssertEventsSaved(
+            //e => Assert.IsType<Events.UserSignedInEvent>(e),
```
gunndabad, line 69: There's a bug currently that prevents a `UserSignedInEvent` from being issued in this situation. I'll fix that separately.

wilpow-github: Is there a better way of testing if the user is signed in/out than using the Events?

wilpow-github (approved):

dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/SignOut.cs
gunndabad: We could put something into the page HTML, I guess. These event checks are still useful, though (we wouldn't have caught this bug without them.)

gunndabad: Fair point. I've added a marker `<span>` to TestClient that indicates logged in-ness and added a check on that to tests.

wilpow-github: looks good to me!

PR #498, opened 2023-04-05
Make update name request to DQT on official name change form submit

wilpow-github:
### Context

We’ve added the DQT change name process but it’s not currently hooked up to the API (that actually creates the incident).

When the final form is submitted call the https://qualified-teachers-api-tech-docs.london.cloudapps.digital/#v3-teachers-name-changes endpoint, passing over the data from the form. The evidenceUrl should be a fully-qualified URL to the blob in blob storage, including a short-lived read-only SAS token.

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Account/OfficialName/Confirm.cshtml.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Account/OfficialName/Confirm.cshtml.cs
@@ -37,9 +37,6 @@ public class Confirm : PageModel
     [FromQuery(Name = "lastName")]
     public string? LastName { get; set; }
 
-    [FromQuery(Name = "fileId")]
-    public string? FileId { get; set; }
-
     [FromQuery(Name = "fileName")]
     public string? FileName { get; set; }
 
@@ -72,7 +69,7 @@ public class Confirm : PageModel
 
     public override void OnPageHandlerExecuting(PageHandlerExecutingContext context)
     {
-        if (FirstName is null || LastName is null || FileId is null || FileName is null)
+        if (FirstName is null || LastName is null || FileName is null)
         {
             context.Result = BadRequest();
         }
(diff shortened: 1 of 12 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/Services/DqtApi/TeacherNameChangeRequest.cs
```diff
@@ -0,0 +1,11 @@
+namespace TeacherIdentity.AuthServer.Services.DqtApi;
+
+public record TeacherNameChangeRequest
```
wilpow-github, line 3: No way to make sure these fields are non-null when using object instantiator, but not a big deal I guess

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Account/OfficialName/Confirm.cshtml.cs
```diff
@@ -38,11 +47,25 @@ public void OnGet()
     {
```
wilpow-github, line 47: Not used:
    [FromQuery(Name = "fileId")]
    public string? FileId { get; set; }

dotnet-authserver/src/TeacherIdentity.AuthServer/Services/DqtApi/DqtApiClient.cs
```diff
@@ -49,4 +51,13 @@ public async Task<GetIttProvidersResponse> GetIttProviders(CancellationToken can
         response.EnsureSuccessStatusCode();
         return (await response.Content.ReadFromJsonAsync<GetIttProvidersResponse>(cancellationToken: cancellationToken))!;
     }
+
+    public async Task PostTeacherNameChange(TeacherNameChangeRequest request, CancellationToken cancellationToken = default)
+    {
+        string json = JsonSerializer.Serialize(request);
+        HttpContent content = new StringContent(json, Encoding.UTF8, "application/json");
```
gunndabad, line 58: Could use `JsonContent.Create()` here

dotnet-authserver/src/TeacherIdentity.AuthServer/Services/DqtApi/TeacherNameChangeRequest.cs
```diff
@@ -0,0 +1,11 @@
...
+{
+    public required string FirstName { get; init; }
+    public string? MiddleName { get; init; }
```
gunndabad, line 6: Even though it's nullable I think forcing this to be explicitly set is no bad thing
```suggestion
    public required string? MiddleName { get; init; }
```

gunndabad (commented):
Do we need to undo any `appsettings.Development.json` here that are currently using the local emulator?

gunndabad (approved):

PR #516, opened 2023-04-12
Mobile number normalization

gunndabad:
Some new end-to-end tests showed that we were missing phone normalization somewhere. In addition, there has been some feedback that we should be able to roundtrip the number the user gave to us, even if when it's not in a normalized form.

This PR ports the phone number validation and normalization code from Notify and bundles it into a new `MobileNumber` type. It's done in a new type so we can pass it around safely and know that's it's been validated. In addition, we add an additional column to `User` - `NormalizedPhoneNumber`. With this we can store both the value the user gave to us (so we can successfully roundtrip without losing data) plus the normalized form (so we can reliably match it).

I've also added a new phone number generation method to `TestData`.


dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/ModelTests/MobileNumberTests.cs
```diff
@@ -0,0 +1,157 @@
...
+                "07\";DROP TABLE;\"",
+                "+44 07ab cde fgh",
+                "ALPHANUM3R1C"
+            }
+        )
+    };
+
+    private static readonly (string ExpectedErrorMessage, string Number)[] _invalidPhoneNumbers =
+        _invalidUkPhoneNumbers
+            .SelectMany(t => t.Numbers.Take(0).Select(n => (t.ExpectedErrorMessage, Number: n)))
```
gunndabad, line 150: This line looks a bit weird but it's mirroring the python tests from Notify.

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Register/ExistingAccountPhoneConfirmation.cshtml.cs
```diff
@@ -25,6 +25,7 @@ public class ExistingAccountPhoneConfirmation : BasePhoneConfirmationPageModel
     }
 
     public override string? MobileNumber => HttpContext.GetAuthenticationState().ExistingAccountMobileNumber;
```
wilpow-github, line 27: I guess technically, this comes from the normalised mobile number column so shouldn't really need parsing again. No harm in being safe though

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Register/PhoneConfirmation.cshtml.cs
```diff
@@ -54,7 +54,7 @@ public async Task<IActionResult> OnPost()
         var permittedUserTypes = authenticationState.GetPermittedUserTypes();
 
         var user = await _dbContext.Users
-            .Where(u => (u.MobileNumber ?? "") == PhoneHelper.FormatMobileNumber(MobileNumber!))
+            .Where(u => u.NormalizedMobileNumber! == parsedMobileNumber)
```
wilpow-github, line 57: Does this work? I was afraid of asserting non-null here

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Register/ExistingAccountPhoneConfirmation.cshtml.cs
gunndabad: This one comes from the un-normalised value but it is assumed to be valid. As this isn't in production yet I figured there's no way an invalid value could ever be observed here.

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Register/PhoneConfirmation.cshtml.cs
gunndabad: It does work, indeed it was the only way I could make the nullable annotations happy (without `#pragma'-ing stuff).

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Register/ExistingAccountPhoneConfirmation.cshtml.cs
wilpow-github: Ah ok, originally was coming from the mobile_number column which was assumed to be normalised. So now it's from the same column but normalisation isn't a guarantee. Got it!

dotnet-authserver/src/TeacherIdentity.AuthServer/Services/UserVerification/UserVerificationService.cs
```diff
@@ -164,8 +164,10 @@ public async Task<PinGenerationResult> GenerateSmsPin(string mobileNumber)
                 });
 
                 // Remove any other active PINs for this email
+#pragma warning disable IDE0071 // Simplify interpolation
                 await _dbContext.Database.ExecuteSqlInterpolatedAsync(
-                    $"update sms_confirmation_pins set is_active = false where mobile_number = {mobileNumber} and expires > {_clock.UtcNow} and is_active = true;");
+                    $"update sms_confirmation_pins set is_active = false where mobile_number = {mobileNumber.ToString()} and expires > {_clock.UtcNow} and is_active = true;");
```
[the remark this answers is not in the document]
gunndabad: C# seems to think the `.ToString()` is redundant but EF's interpolated string magic _really_ doesn't like it. The converter added to the `DbContext` doesn't seem to be used with `ExecuteSqlInterpolatedAsync`.

dotnet-authserver/tests/TeacherIdentity.AuthServer.TestCommon/TestData.CreateUser.cs
```diff
@@ -15,7 +14,7 @@ public partial class TestData
             TrnLookupStatus? trnLookupStatus = null,
             string? firstName = null,
             string[]? staffRoles = null,
-            bool? hasMobileNumber = true) =>
+            bool? hasMobileNumber = null) =>
```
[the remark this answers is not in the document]
gunndabad: I changed it to `null` so that `Staff` users wouldn't get assigned a mobile by default (since they never will IRL). There's a line further down that sets `hasMobileNumber` to `true` if `null` were specified and `UserType == Default`.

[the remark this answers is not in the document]
wilpow-github: Ah missed that!

wilpow-github (approved):

PR #518, opened 2023-04-12
Use a fixed verification PIN for dev environments

gunndabad:
For local development and some non-production environments it would be useful to have a known, fixed PIN for email & SMS verification to remove the need for picking through logs to find a 'real' code.

This change adds a `FixedPinUserVerificationService` implementation of `IUserVerificationService` and enables it when a `UserVerification:UseFixedPin` config entry is set. It's enabled for local dev by default and also our end-to-end tests.

Developers can still use the real implementation if desired by overriding the config value in user secrets.

### DIFF
--- a/dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/HostFixture.cs
+++ b/dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/HostFixture.cs
@@ -12,7 +12,6 @@ using TeacherIdentity.AuthServer.EventProcessing;
 using TeacherIdentity.AuthServer.Models;
 using TeacherIdentity.AuthServer.Oidc;
 using TeacherIdentity.AuthServer.Services.DqtApi;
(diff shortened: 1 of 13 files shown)


dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/HostFixture.cs
```diff
@@ -31,22 +30,19 @@ public class HostFixture : IAsyncLifetime
     private bool _disposed = false;
 
     private readonly List<string> _capturedAccessTokens = new();
-    private readonly List<(string Email, string Pin)> _capturedEmailConfirmationPins = new();
+
+    public static string UserVerificationPin { get; } = "12345";
```
gunndabad, line 34: This is set statically so that extension methods (e.g. on `PageExtensions`) could access it directly without having to accept a parameter.

hortha (approved):

PR #522, opened 2023-04-13
Use Notify for generating email and SMS templates

gunndabad:
We want to move the templates into Notify itself so that content designers can change them without developer involvement. In addition we will be able to use Notify’s ‘hide personalisation’ feature.

### DIFF
--- a/terraform/app.tf
+++ b/terraform/app.tf
@@ -33,9 +33,9 @@ locals {
       EncryptionKeys__1                            = local.infrastructure_secrets.ENCRYPTION_KEY1,
       SigningKeys__0                               = local.infrastructure_secrets.SIGNING_KEY0,
       SigningKeys__1                               = local.infrastructure_secrets.SIGNING_KEY1,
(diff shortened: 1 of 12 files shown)


terraform/app.tf
```diff
@@ -33,9 +33,9 @@ locals {
       EncryptionKeys__1                            = local.infrastructure_secrets.ENCRYPTION_KEY1,
       SigningKeys__0                               = local.infrastructure_secrets.SIGNING_KEY0,
       SigningKeys__1                               = local.infrastructure_secrets.SIGNING_KEY1,
-      Notify__ApiKey                               = local.infrastructure_secrets.NOTIFY_API_KEY,
+      Notify__ApiKey                               = local.infrastructure_secrets.NOTIFY_ID_API_KEY,
```
gunndabad, line 36: Keys have changed so we can roll this out more easily, since we're also moving to a new Notify service.

gunndabad:
[Move confirmation templates into Notify](https://trello.com/c/sn3KPjP1/915-move-confirmation-templates-into-notify)

hortha (approved):

PR #528, opened 2023-04-16
Sign in journey abstraction

gunndabad:
Today there are many places where we handle sign in journey orchestration; `AuthenticationState` holds all state for the journey and has _some_ routing knowledge. Individual pages, for the most part, know where to go next and where to go back to. On top of this we have a variety of filters - `RequireAuthenticationMilestoneAttribute`, `CheckUserRequirementsForTrnJourneyFilterFactory` and many overrides of `OnPageHandlerExecuting()` in individual pages that check that the user is allowed to access the current page. These filters aren't super consistent either; sometimes we redirect users to an available page and sometimes we return a `400` error.

It would be nice to move this orchestration logic to somewhere central so that we can get more consistency and remove repetition.

This PR proposes an abstraction for managing sign in journeys - `SignInJourney`. There's also a `SignInJourneyProvider` which gets a `SignInJourney` given an `AuthenticationState`. I've added stub classes for the current journeys we support too.

Each journey implements its own class inheriting from `SignInJourney`. This class exposes the following capabilities:
- the start step for the journey;
- whether the journey is finished;
- the next step in the journey, given the current page;
- the previous step in the journey, given the current page;
- whether a given step is accessible given the state of the journey;
- the last accessible step given the state of the journey.

With this we can simplify a lot:
- All the filters mentioned above and `OnPageHandlerExecuting()` implementations that check if the current page is accessible can be replaced with a single attribute - `CheckCanAccessStepAttribute` - that delegates to the journey's `CanAccessStep` method. This will work even if the page is used in multiple journeys and the logic for its availability differs between journeys.
- Logic that is common over multiple journeys can be put into the appropriate base class (e.g. creating a user, looking up a TRN).

The following types/methods could ultimately get removed if we rolled this out everywhere:
- `AuthenticationJourneyType`
- `AuthenticationMilestone`
- `AuthenticationState.GetNextHopUrl()`
- `AuthenticationState.IsComplete()`

Ultimately it would be good to split up `AuthenticationState` into journey-specific state types as well and this provides some groundwork for doing that.

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/SignInJourney.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/SignInJourney.cs
@@ -23,33 +23,15 @@ public abstract class SignInJourney
 
     public abstract bool IsFinished();
 
-    public virtual string GetStartStep() => Steps.Email;
+    public abstract string GetStartStep();
 
-    public virtual string GetStepUrl(string step) => step switch
-    {
-        Steps.Email => LinkGenerator.Email(),
-        Steps.EmailConfirmation => LinkGenerator.EmailConfirmation(),
-        _ => throw new ArgumentException($"Unknown step: '{step}'.")
-    };
+    public abstract string GetStepUrl(string step);
 
-    public virtual string? GetNextStep(string currentStep) => currentStep switch
-    {
-        Steps.Email => Steps.EmailConfirmation,
-        _ => null
-    };
+    public abstract string? GetNextStep(string currentStep);
 
-    public virtual string? GetPreviousStep(string currentStep) => currentStep switch
-    {
-        Steps.EmailConfirmation => Steps.Email,
-        _ => null
-    };
+    public abstract string? GetPreviousStep(string currentStep);
 
-    public virtual bool CanAccessStep(string step) => step switch
-    {
-        Steps.Email => !AuthenticationState.EmailAddressVerified,
-        Steps.EmailConfirmation => !AuthenticationState.EmailAddressVerified && AuthenticationState.EmailAddressSet,
-        _ => false
-    };
+    public abstract bool CanAccessStep(string step);
 
     public virtual string GetLastAccessibleStep()
     {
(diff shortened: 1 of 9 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/SignInJourney.cs
```diff
@@ -0,0 +1,141 @@
...
+        {
+            throw new InvalidOperationException("Email cannot be verified at this time.");
+        }
+
+        return OnEmailVerifiedCore(user);
+    }
+
+    protected abstract Task<IActionResult> OnEmailVerifiedCore(User? user);
+
+    public static class Steps
```
gunndabad, line 136: I've made each step a `string`. I experimented with enums but found them a little more difficult to use in practice (with `const` fields and attributes in particular).

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/SignInJourney.cs
```diff
@@ -0,0 +1,141 @@
...
+
+    public HttpContext HttpContext { get; }
+
+    public IdentityLinkGenerator LinkGenerator { get; }
+
+    public abstract bool IsFinished();
+
+    public virtual string GetStartStep() => Steps.Email;
+
+    public virtual string GetStepUrl(string step) => step switch
```
gunndabad, line 28: The `GetStepUrl`, `GetNextStep`, `GetPreviousStep` and `CanAccessStep` methods are expected to overriden by deriving types. The idea is that deriving types implement the logic for the pages their journey-specific pages then delegate to the `base` implementation for the shared pages (e.g. `Email`, `EmailConfirmation`).

Another approach would be to make these all abstract and have each deriving type implement the method completely.

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/SignInJourney.cs
wilpow-github: Did you consider making each step a class with NextStep, PreviousStep, stepRequirements, etc. member properties? Might be useful when there is e.g. tricky logic to determine the next step

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/SignInJourney.cs
gunndabad: That makes sense. In reality the base class can't completely encapsulate the `Email` + `EmailConfirmation` steps any way because it won't know what comes next.

PR #530, opened 2023-04-17
Change official name refinements

gunndabad:
Mirrors the refinements as done for Change official name.

- Initialize the date on the details page to what the user entered when coming from the Confirm page.
- Redirect back to the Confirm page when clicking Continue on the Details page when the user has come from the Confirm page.
- Use the name of the uploaded file on the Confirm page rather than our generated name.


gunndabad:
[Change DQT name/DOB snags](https://trello.com/c/4m1sqWQ7/918-change-dqt-name-dob-snags)

hortha (approved):

PR #533, opened 2023-04-18
Fix date format to remove leading '0' from day

gunndabad:


gunndabad:
[Dates should not have leading zeros](https://trello.com/c/wyYV8K3Y/925-dates-should-not-have-leading-zeros)

hortha (approved):

PR #534, opened 2023-04-18
Add middle name to Official name row on account page

gunndabad:


dotnet-authserver/src/TeacherIdentity.AuthServer/Services/DqtApi/TeacherInfo.cs
```diff
@@ -5,7 +5,7 @@ public record TeacherInfo
     public required string Trn { get; init; }
     public required string FirstName { get; init; }
     public required string LastName { get; init; }
-    public required string? MiddleName { get; init; }
+    public required string MiddleName { get; init; }
```
gunndabad, line 8: API now guarantees this will never be `null`.

hortha (approved):

PR #550, opened 2023-04-21
Always infer Pending if Find API returns multiple results

gunndabad:
We currently treat multiple results from TRN lookups the same as no results. This changes the resolution logic so that we never resolve to `TrnLookupStatus.None` if there are any results, even if the user's answers suggest that they probably don't have a TRN.


gunndabad:
[Never derive TrnLookupStatus of `None` if there are multiple matches](https://trello.com/c/gRyxewoA/936-never-derive-trnlookupstatus-of-none-if-there-are-multiple-matches)

wilpow-github (approved):

PR #554, opened 2023-04-25
Create recurring job to purge expired confirmation pins

wilpow-github:
### Context

We stash PINs for email and SMS verification in our DB but they’re never removed.

Create a recurring Hangfire job that purges old rows from the email_confirmation_pins and sms_confirmation_pins tables. And old row is any row where the code expired at least 7 days ago.

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Jobs/PurgeConfirmationPinsJob.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Jobs/PurgeConfirmationPinsJob.cs
@@ -1,3 +1,4 @@
+using Microsoft.EntityFrameworkCore;
 using TeacherIdentity.AuthServer.Models;
 
 namespace TeacherIdentity.AuthServer.Jobs;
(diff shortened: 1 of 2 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/Jobs/PurgeConfirmationPinsJob.cs
```diff
@@ -0,0 +1,47 @@
...
+    private async Task DeleteSmsConfirmationPins(CancellationToken cancellationToken)
+    {
+        var expiredSmsConfirmationPins = _dbContext.SmsConfirmationPins
+            .Where(p => p.Expires < DateTime.UtcNow - _expiredPinGracePeriod)
+            .ToList();
+
+        if (expiredSmsConfirmationPins.Any())
+        {
+            _dbContext.SmsConfirmationPins.RemoveRange(expiredSmsConfirmationPins);
+            await _dbContext.SaveChangesAsync(cancellationToken);
```
wilpow-github, line 44: I had some invalid mobile numbers in my DB from before the international prefix was added which caused this transaction to roll back. Do we want to make this into multiple transactions or are we happy with this?

gunndabad: Could we do something like https://github.com/DFE-Digital/get-an-identity/blob/2891711a053b1cc7db9b055e4a5922e2404bee1a/dotnet-authserver/src/TeacherIdentity.AuthServer/Services/UserVerification/UserVerificationService.cs#L80 to avoid even pulling the records back at all?

gunndabad (approved):

PR #562, opened 2023-05-02
Add preferred_name claim

gunndabad:
In the future we're going to add a separate preferred name field for ID accounts. We'll be sending this as an additional optional claim. So that services can begin to prepare for this claim being added, add it now but populate it with `name`.

I've also fixed up the naming format for our custom claims; I've checked and nobody externally is depending on these names just yet.

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/CustomClaims.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/CustomClaims.cs
@@ -5,8 +5,8 @@ public static class CustomClaims
     public const string DateFormat = "yyyy-MM-dd";
 
     public const string Trn = "trn";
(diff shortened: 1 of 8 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/Oidc/CustomClaims.cs
```diff
@@ -9,4 +9,5 @@ public static class CustomClaims
     public const string HaveCompletedTrnLookup = "completed-trn-lookup";
     public const string UserType = "user-type";
     public const string PreviousUserId = "previous-user-id";
+    public const string PreferredName = "preferred-name";
```
gunndabad, line 12: Used a `-` to separate words for consistency even though `_` is used in 'official' claims. Changing the claims above now would be a breaking change.

PR #578, opened 2023-05-09
Update core sign in journey to better handle updating mobile number

wilpow-github:
### Context

In the core journey we allow a user to go back to the phone page to change their number from the Check answers page. If they were to change their number, but not verify it, then go back to Check answers (using the browser’s back button) they could create an account with an unverified phone number.

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master


wilpow-github:
The ticket asks for adding an UnverifiedMobileNumber property to authentication state, which would lead to a lot of changes updating instances of MobileNumber to UnverifiedMobileNumber.

Using the back button from /register/phone after making a change to MobileNumber caused an exception to be thrown because the journey attempted to load /register/date-of-birth/ as the latest accessible step. These changes mean that the back button from phone renders phone-confirmation. 

Happy to go with initial suggestion if there's something I'm missing though!

gunndabad (approved):

PR #583, opened 2023-05-10
Update to use EntityFramework Exceptions nuget package

wilpow-github:
### Context

There’s a library that gives more-specific exceptions when database errors occur. By using this we could clean up some our exception handling code.

Replace our catch (DbUpdateException) code with the specific exception type. 

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/ExceptionExtensions.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/ExceptionExtensions.cs
@@ -1,12 +1,11 @@
-using Microsoft.EntityFrameworkCore;
+using EntityFramework.Exceptions.Common;
 using Npgsql;
 
 namespace TeacherIdentity.AuthServer;
 
 public static class ExceptionExtensions
 {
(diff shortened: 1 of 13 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/ExceptionExtensions.cs
```diff
@@ -1,12 +1,11 @@
-using Microsoft.EntityFrameworkCore;
+using EntityFramework.Exceptions.Common;
 using Npgsql;
 
 namespace TeacherIdentity.AuthServer;
 
 public static class ExceptionExtensions
 {
-    public static bool IsUniqueIndexViolation(this DbUpdateException ex, string indexName) =>
+    public static bool IsUniqueIndexViolation(this UniqueConstraintException ex, string indexName) =>
```
wilpow-github, line 8: This still seemed to be the best way to capture the specific exception thrown

wilpow-github:
I've only changed instances of DbUpdateException to UniqueConstraintException here, there are still cases where we do not capture any exception at all and perhaps should be

PR #591, opened 2023-05-12
Add register complete holding page

wilpow-github:
### Context

Until now we’ve not been raising support tickets when we can’t automatically find a user’s TRN. When these users return, they would be stuck if they tried to enter a service that requires a TRN since their TrnLookupStatus will never move on from Pending.

Amend the /SignIn/Complete page for returning users whose TrnLookupStatus is Pending and TrnLookupSupportTicketCreated is false and the client’s TrnRequirementType is Required.

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Complete.cshtml.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Complete.cshtml.cs
@@ -75,7 +75,7 @@ public class CompleteModel : PageModel
         TrnLookupStatus = authenticationState.TrnLookupStatus;
         TrnRequirementType = authenticationState.OAuthState?.TrnRequirementType;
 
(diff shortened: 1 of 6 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/SignIn/Complete.cshtml.cs
```diff
@@ -67,6 +75,9 @@ public async Task OnGet()
         TrnLookupStatus = authenticationState.TrnLookupStatus;
         TrnRequirementType = authenticationState.OAuthState?.TrnRequirementType;
 
+        var user = await _dbContext.Users.SingleOrDefaultAsync(u => u.UserId == authenticationState.UserId);
```
gunndabad, line 78:
```suggestion
        var user = await _dbContext.Users.SingleAsync(u => u.UserId == authenticationState.UserId);
```

gunndabad (approved):

PR #593, opened 2023-05-15
Add trn lookup support ticket created flag to user

hortha:
### Context

When users have gone through ID’s sign up process, if we couldn’t find a TRN then in most cases the user’s TrnLookupStatus is set to Pending. The intention is that these are resolved via the helpdesk manually (following a Zendesk ticket being raised) and ultimately transition to either Failed or Found. However, since NPQ can’t actually handle receiving users without a TRN they are redirecting users at TrnLookupStatus Foundthrough their old journey. Consequently, we haven’t been creating Zendesk tickets and so no one has looked at these records and they’re all stuck inPending. That’s going to be a problem when these users come back to ID and try to access a service that requires a TRN like Access your teaching qualifications.

Add a TrnLookupSupportTicketCreated property to User, defaulting to false. Modify the CreateUserHelper to populate this field. If a Zendesk ticket is created, set the flag to true, otherwise false.

Add a TrnLookupSupportTicketCreatedEvent and emit this when a Zendesk ticket is created too.

Modify the admin user list page. Add an additional checkbox filters within the Pending TRN lookup status
"With support ticktet". 
This should default to ‘unchecked’.

### Guidance to review

Include any useful information needed to review this change.
Include any dependencies that are required for this change.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/EndpointTests/SignIn/Register/CheckAnswersTests.cs
+++ b/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/EndpointTests/SignIn/Register/CheckAnswersTests.cs
@@ -171,13 +171,14 @@ public class CheckAnswersTests : TestBase
 
         if (expectZendeskTicketCreated)
         {
-            elementInspectors.Add(e =>
-            {
-                var supportTicketCreatedEvent = Assert.IsType<TrnLookupSupportTicketCreatedEvent>(e);
-                Assert.Equal(ticketIdExpected, supportTicketCreatedEvent.TicketId);
-                Assert.Equal(Clock.UtcNow, supportTicketCreatedEvent.CreatedUtc);
-                Assert.Equal(user?.UserId, supportTicketCreatedEvent.UserId);
-            });
+            elementInspectors.Add(
+                e =>
+                {
+                    var supportTicketCreatedEvent = Assert.IsType<TrnLookupSupportTicketCreatedEvent>(e);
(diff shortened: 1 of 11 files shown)


hortha:
[Add TrnLookupSupportTicketCreated flag to User](https://trello.com/c/GUZQuTnI/952-add-trnlookupsupportticketcreated-flag-to-user)

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/EndpointTests/SignIn/Register/CheckAnswersTests.cs
```diff
@@ -128,6 +130,16 @@ public async Task Post_JourneyHasExpired_RendersErrorPage()
             authState.OnTrnLookupCompleted(trn, trnLookupStatus);
         }
 
+        TicketCreateRequest? ticketCreateRequestActual = null;
+        long ticketIdExpected = 1234567;
```
gunndabad, line 134: Nit: this looks like a TRN, could we alter it a little to make it super clear?

gunndabad (approved):

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/EndpointTests/SignIn/Register/CheckAnswersTests.cs
hortha: Sure

PR #594, opened 2023-05-15
Amended so that zendesk ticket creation is done in a hangfire job

hortha:
### Context

Today our Zendesk API call to create a ticket happens synchronously around user creation. If that were to fail, our DB would have the new user record in but the user would see an error, caused by the exception. We should put the API call into a job so it can fail and retry automatically.

Move the Zendesk ticket creation step into a Hangfire job.

### Guidance to review

N/A

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master


hortha:
[Run Zendesk ticket creation in a Hangfire job](https://trello.com/c/K1b2BJy6/1001-run-zendesk-ticket-creation-in-a-hangfire-job)

gunndabad (approved):

PR #596, opened 2023-05-16
Add Landing page for TRN token sign in journey

wilpow-github:
### Context

We have a revised journey for those registering with a magic link. This is the initial landing page for that journey.

Create a new page at /sign-in/trn-token/ following the designs at https://get-an-identity-prototype.herokuapp.com/auth/ga-account . Access your teaching qualifications should be dynamic and use the current client’s display name.

Create a new journey - TrnTokenRegistrationJourney - that is used when a valid trn_token is provided in the authorize endpoint (we may need to add a flag to AuthenticationState so we know that).

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/AuthenticationState.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/AuthenticationState.cs
@@ -60,6 +60,7 @@ public class AuthenticationState
         UserType = authStateInitData?.UserType;
         StaffRoles = authStateInitData?.StaffRoles;
         TrnLookupStatus = authStateInitData?.TrnLookupStatus;
+        TrnToken = authStateInitData?.TrnToken;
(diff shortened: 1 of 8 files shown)


dotnet-authserver/src/TeacherIdentity.AuthServer/AuthenticationState.cs
```diff
@@ -139,6 +140,8 @@ public class AuthenticationState
     public string? ExistingAccountMobileNumber { get; private set; }
     [JsonInclude]
     public bool? ExistingAccountChosen { get; private set; }
+    [JsonInclude]
+    public bool? TrnToken { get; private set; }
```
gunndabad, line 144: Does this need to be nullable?

gunndabad (approved):

PR #597, opened 2023-05-17
Trn token phone pages

wilpow-github:
### Context

We have a revised journey for those registering with a magic link. This is the phone number & phone confirmation pages for that journey.

Amend the /sign-in/register/phone and /sign-in/register/phone-confirmation pages such that they work with the new magic link journey.

Amend the journey so that the ‘Register’ link from the landing page navigates to /sign-in/register/phone.

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


wilpow-github:
Might need to think about what to do for unit tests. We'd only be checking potentially different redirects, so perhaps can be covered by e2e tests?

gunndabad:
> Might need to think about what to do for unit tests. We'd only be checking potentially different redirects, so perhaps can be covered by e2e tests?

Yes, e2e tests are enough for me.

gunndabad (approved):

PR #601, opened 2023-05-18
Additional zendesk ticket info

aje54:
### Context

When the helpdesk are manually resolving a TRN they need to know whether the user was prevented from signing in so that they can email them once their TRN has been found. For NPQ, the user would not be emailed since a TRN is not required to sign in to NPQ. For the new Quals UI, the user would be prevented from signing in since the Quals UI requires a TRN to retrieve the user’s data.

Add an additional two lines to the ticket content:

Service: {service name}
Contact user with TRN: {Yes or No}
Contact user with TRN should be Yes if the current client’s TrnRequirementType is Required, otherwise No.

### Guidance to review

Include any useful information needed to review this change.
Include any dependencies that are required for this change.

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/CoreSignInJourneyWithTrnLookup.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/CoreSignInJourneyWithTrnLookup.cs
@@ -1,5 +1,4 @@
 using EntityFramework.Exceptions.Common;
-using Microsoft.AspNetCore.Components.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using TeacherIdentity.AuthServer.Oidc;
 using TeacherIdentity.AuthServer.Services.BackgroundJobs;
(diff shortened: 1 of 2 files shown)


hortha (approved):

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/CoreSignInJourneyWithTrnLookup.cs
```diff
@@ -53,7 +53,9 @@ public override async Task<IActionResult> CreateUser(string currentStep)
                             AuthenticationState.DateOfBirth,
                             AuthenticationState.NationalInsuranceNumber,
                             AuthenticationState.IttProviderName,
-                            AuthenticationState.StatedTrn));
+                            AuthenticationState.StatedTrn,
+                            AuthenticationState.OAuthState.ClientId,
```
hortha, line 57: This needs to be the client "display name"

hortha (changes requested):

hortha (approved):

PR #610, opened 2023-05-24
Update trn token journey

wilpow-github:
### Context

We’ve largely built the ‘happy path’ of a user arriving at ID via a magic link and registering a new account. There a quite a few edge cases; those are covered here.

When a user arrives at the authorize endpoint with a valid TRN and a user is already signed in:

when the signed in user’s email address matches that of the TRN token:

when the existing account’s TRN matches the token’s TRN, sign the user in and invalidate the token (by setting the token’s UserId to that of the existing user) and redirect to the complete page (i.e. the user shouldn’t see the landing page);

when the account does not have a TRN, assign the TRN from the TRN token (and set the TrnAssociationSource to TrnToken and set the token’s UserId) and sign the user in;

when the existing account has TRN that does not match the TRN of the token, discard the token and fall back to a regular core journey.

when the signed in user’s email address does not match that of the TRN token, reset AuthenticationState and populate from the TRN token, as if the user were not signed in at all.

When a user arrives at the authorize endpoint with a valid TRN token and that token's email address matches an existing account:

when the existing account’s TRN matches the token’s TRN, sign the user in and invalidate the token (by setting the token’s UserId to that of the existing user) and redirect to the complete page (i.e. the user shouldn’t see the landing page);

when the existing account does not have a TRN, assign the TRN from the TRN token (and set the TrnAssociationSource to TrnToken and set the token’s UserId);

when the existing account has a TRN that does not match the TRN of the token, discard the token and fall back to a regular core journey.

When a user arrived at the authorize endpoint with a valid TRN token and they 'sign in' from the landing page into an existing account:

when that account does not have a TRN, assign the TRN from the TRN token (and set the TrnAssociationSource to TrnToken and set the token’s UserId);

when that account has a different TRN to that of the TRN token then do not overwrite the TRN of the existing account; sign them in as if there were no TRN token;

when that account has the same TRN as the TRN token, sign them in as if there were no TRN token (we should have this covered by the up-front check that there’s no existing account for the TRN of the token)

When beginning the register journey, if there is already an account that matches the pre-populated name and DOB then show the /sign-in/register/account-exists page. If a user chooses to sign in to the existing account and they verify successfully and that account does not have a TRN, assign the TRN from the token to that account (and set the TrnAssociationSource to TrnToken and set the token’s UserId). If the existing account already has a TRN, invalidate the TRN token.

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master


dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/TrnTokenSignInJourney.cs
```diff
@@ -110,4 +114,524 @@ public async Task NewTeacherUser_WithTrnTokenChangesEmail_CreatesUserAndComplete
...
+        await page.SubmitCompletePageForExistingUser();
+
+        await page.AssertSignedInOnTestClient(user.EmailAddress, user.Trn, user.FirstName, user.LastName);
+
+        var trnTokenModel = await _hostFixture.TestData.GetTrnToken(trnToken.TrnToken);
+        Assert.Null(trnTokenModel?.UserId);
+    }
+
+    [Fact]
+    public async Task ExistingUserSignsIn_ValidTrnTokenMatchingTrn_SignInAndInvalidatesToken()
```
wilpow-github, line 392: This Invalidates the token rather than Ignores it. Follows the pattern of invalidating tokens when users sign in with same Trn

wilpow-github:
The logic covering these edge cases can be difficult to understand and is quite fragile. It's been through a number of iterations and this is the closest I've got to satisfying the ACs in a reasonable way. Happy to chat if it's not quite what we want.

dotnet-authserver/src/TeacherIdentity.AuthServer/Controllers/AuthorizationController.cs
```diff
@@ -126,13 +113,23 @@ public async Task<IActionResult> Authorize()
...
                     TrnRequirementType = trnRequirementType
-                });
+                },
+                authenticateResult.Succeeded != true);
+
+            var signedInUser = await GetSignedInUser(authenticateResult, userRequirements);
+
+            if (trnToken is not null)
+            {
+                await _trnTokenHelper.InitializeAuthenticationStateWithToken(signedInUser, authenticationState, trnToken, HttpContext);
```
wilpow-github, line 127: What is happening here:
If there is no signed in user, but we find an existing user with an email match, we add the existing users details to auth state. When we hit lines 189, 190
` var signInJourney = _signInJourneyProvider.GetSignInJourney(authenticationState, HttpContext);
            return Redirect(signInJourney.GetStartStepUrl());`
We get redirected to this endpoint again, except this time we DO have a signed in user. I don't know why 
`return Redirect(signInJourney.GetStartStepUrl())` ends up back here, so this is the best solution I've come up with so far

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/TrnTokenHelper.cs
```diff
@@ -0,0 +1,171 @@
...
+                await authenticationState.SignIn(httpContext);
+            }
+            else
+            {
+                var users = await _userSearchService.FindUsers(trnToken.FirstName, trnToken.LastName, trnToken.DateOfBirth);
+                authenticationState.OnExistingAccountSearch(users.Length == 0 ? null : users[0]);
+                authenticationState.OnTrnTokenProvided(trnToken);
+            }
+        }
+    }
```
wilpow-github, line 92: I think we can sacrifice readability to reduce nesting here, but not sure

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/TrnTokenHelper.cs
```diff
@@ -0,0 +1,171 @@
...
+        {
+            user = await UpdateUserTrn(user, trnToken.Trn);
+        }
+
+        if (user.Trn == trnToken.Trn)
+        {
+            await InvalidateTrnToken(trnToken.TrnToken, user.UserId);
+        }
+
+        return user;
```
wilpow-github, line 109: This is the main reason why the ACs on a couple of tests are don't match the ticket. If we were to implement as requested I don't think this function could exist, we would have to invalidate tokens on a case by case basis (and probably add some parameters to determine the current edge case we are trying to handle)

dotnet-authserver/src/TeacherIdentity.AuthServer/Controllers/AuthorizationController.cs
wilpow-github, line 1: There is an issue in that this endpoint is hit again after the user is signed in. So it is hard to remain consistent with the trn token logic on both passes of this logic

dotnet-authserver/src/TeacherIdentity.AuthServer/Controllers/AuthorizationController.cs
```diff
@@ -184,15 +186,17 @@ public async Task<IActionResult> Authorize()
...
-
             var signInJourney = _signInJourneyProvider.GetSignInJourney(authenticationState, HttpContext);
             return Redirect(signInJourney.GetStartStepUrl());
         }
 
+        if (trnToken is not null)
+        {
+            // If we have a signed in user we may never enter the TRN token sign in journey, so we have to apply
+            // the token here
+            await _trnTokenHelper.ApplyTrnTokenToUser(authenticationState.UserId, trnToken.TrnToken);
```
wilpow-github, line 197: This now looks wrong to me. There are no checks between user Email and Trn Token Email for example

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/TrnTokenHelper.cs
```diff
@@ -0,0 +1,171 @@
...
+            LastName = teacher.LastName,
+            DateOfBirth = teacher.DateOfBirth,
+        };
+    }
+
+    private async Task<User> UpdateUserTrn(User user, string trn)
+    {
+        user.Trn = trn;
+        user.TrnLookupStatus = TrnLookupStatus.Found;
+        user.CompletedTrnLookup = null;
```
gunndabad, line 153: Not sure we want to necessarily `null` this out; if the user has gone through the lookup before (but not found a TRN) then I think we'd want to keep the original timestamp.

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/TrnTokenHelper.cs
gunndabad: I think it's fine as-is.

dotnet-authserver/src/TeacherIdentity.AuthServer/Controllers/AuthorizationController.cs
gunndabad: We could remove the `trn_token` parameter for the callback (see `GetCallbackUrl`) to avoid logic running twice.

dotnet-authserver/tests/TeacherIdentity.AuthServer.EndToEndTests/TrnTokenSignInJourney.cs
```diff
@@ -110,4 +114,524 @@ public async Task NewTeacherUser_WithTrnTokenChangesEmail_CreatesUserAndComplete
...
+
+        await page.SubmitCompletePageForExistingUser();
+
+        await page.AssertSignedInOnTestClient(user, expectTrn: true);
+
+        var trnTokenModel = await _hostFixture.TestData.GetTrnToken(trnToken.TrnToken);
+        Assert.Equal(user.UserId, trnTokenModel?.UserId);
+
+        _hostFixture.EventObserver.AssertEventsSaved(
+            e => Assert.IsType<UserSignedInEvent>(e),
```
gunndabad, line 140: Can we fix the double event here?

dotnet-authserver/src/TeacherIdentity.AuthServer/Controllers/AuthorizationController.cs
```diff
@@ -126,13 +114,16 @@ public async Task<IActionResult> Authorize()
...
                 sessionId,
                 oAuthState: new OAuthAuthorizationState(request.ClientId!, request.Scope!, request.RedirectUri)
                 {
                     TrnRequirementType = trnRequirementType
-                });
+                },
+                authenticateResult.Succeeded != true);
+
+            var signedInUser = await GetSignedInUser(authenticateResult, userRequirements);
+            var existingUserClaimsPrincipal = await InitializeAuthenticationState(signedInUser, trnToken, authenticationState);
```
wilpow-github, line 125: This is a bit of an awkward way to retrieve the claims principal for the new signed in user, perhaps there's a better way of doing it?

gunndabad: Seems reasonable to me and I can't think of anything better.

gunndabad (approved):

gunndabad (approved):

gunndabad (approved):

PR #613, opened 2023-05-25
Generate lower case TRN tokens

gunndabad:
Upper case TRN tokens look a bit shouty when appended to a URL so this changes them to be lower case.

I've tweaked the name of the request & response classes to align with the naming we use elsewhere too.


dotnet-authserver/src/TeacherIdentity.AuthServer/Api/V1/Validators/PostTrnTokensRequestValidator.cs
```diff
@@ -4,7 +4,7 @@
 
 namespace TeacherIdentity.AuthServer.Api.V1.Validators;
 
-public class PostTrnTokensRequestValidator : AbstractValidator<PostTrnTokensRequest>
+public class PostTrnTokensRequestValidator : AbstractValidator<CreateTrnTokenRequest>
```
hortha, line 7: Should this be called CreateTrnTokenRequestValidator?

hortha (changes requested):

hortha (approved):

hortha (approved):

PR #618, opened 2023-05-30
Add feature to change middle name from account page

wilpow-github:
### Context

We’ve added the MiddleName field to User, now we need a way for existing users to change it.

Add an optional MiddleName field to /account/name with label ‘Enter your middle name’. If the middle name is over 200 characters, show an error ‘Middle name must be 200 characters or less’.

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


wilpow-github:
Added both of you as I think the review is rather straightforward

hortha (approved):

PR #619, opened 2023-06-01
Add new institution email page to register journey

wilpow-github:
### Context

Until recently we had a hard block on institution email addresses. We had to remove that as it’s a little too aggressive. In its place we’re adding a page to suggest using a personal email address and it’s on the user to decide whether to use the email or not.

For the core and TRN token journeys, after the email address has been verified add a check that the user’s email is an institution address (unless we have an existing account with the email address). If it is, show an interstitial page matching the design at https://dfe-identity-account-prototype.herokuapp.com/auth/institution-email .

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/CoreSignInJourney.cs
wilpow-github, line 1: It's not very clear what is going on in the sign-in journey pages. The basic idea is we specify whether a page can be accessed based on the authentication state, what the next and previous pages in the journey are.

This is complicated by the fact that e.g. from the check answers page the user should be able to change the email and be directed straight back to check-answers without having to go through all the registration pages again

hortha (approved):

PR #620, opened 2023-06-02
Add preferred name to user model, events and notifications

wilpow-github:
### Context

We want a separate, optional preferred name field on a user’s account so that a user can specify an alternative name to be used in correspondence to what’s held in their DQT record.

Add a PreferredName field to User and a corresponding database column with max length 200.

### Checklist

-   [ ] Attach to Trello card
-   [ ] Rebased master


docs/webhooks.md
wilpow-github, line 1: middleName should have been added when middleName was added to user

docs/webhooks.md
```diff
@@ -68,6 +72,7 @@ The `user` object contains the complete set of user information. `trn` may be `n
   "changes": {
     "emailAddress": "",
     "firstName": "",
+    "middleName": "",
     "lastName": "",
```
wilpow-github, line 76: Not adding preferredName to changes yet as was stated in the ticket

dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/CreateUserHelper.cs
```diff
@@ -87,6 +87,11 @@ public async Task<User> CreateUserWithTrnLookup(AuthenticationState authenticati
             TrnLookupStatus = authenticationState.TrnLookupStatus
         };
 
+        if (populatePreferredName)
```
wilpow-github, line 90: I do wonder if we need to restrict assigning preferred name to only the legacy TRN journey. Would be nice to not have to do this check

hortha: this seems to imply it will be available in other journeys
https://trello.com/c/SFWbP5eY/1041-add-preferred-name-to-core-registration-journey

hortha (approved):

PR #625, opened 2023-06-07
Remove the journey timeout

gunndabad:
20 minutes is too short a time to require users to complete registration within. This removes the timeout altogether for now.

I've kept most of the code around that references journey expiration in case we need to add this back in future.


gunndabad:

wilpow-github (approved):

PR #632, opened 2023-06-09
Don't use the service name in the header except for Legacy TRN journey

gunndabad:
Historically we've used the service name in our header so that ID is inseparable from the calling service. We no longer want to be doing that. This change amends the header so that we always use 'DfE Identity account', except for the Legacy TRN journey.


gunndabad:

wilpow-github (approved):

PR #636, opened 2023-06-12
Show normalized phone number on Check answers page

gunndabad:
Adds a `ToDisplayString()` method to `MobileNumber` which returns a normalized version of the phone number for display purposes. (The normalized form we actually store isn't great for displaying directly.) We special case the UK country code so we display a leading '0' instead of '+44'.


wilpow-github:
Should we update TrnToken/CheckAnswers?

PR #651, opened 2023-07-03
Add end to end tests for webhooks

hortha:
### Context

We’re adding more logic to our webhook dispatcher and we don’t currently have tests that cover actually delivering a webhook.

We want tests that call INotificationPublisher.PublishNotification then assert that a particular JSON message was received over HTTP and that the signature was correct. We should cover all message types and include logic around message type filtering.

Add a new set of tests under Tests.WebHooks with a new TestBase plus a new fixture that derives from HostFixture. This new fixture should add a new endpoint that can receive the webhook messages (e.g. /test-webhook ) and add a new WebHook into the database that matches its address. The endpoint itself should stash away all messages (as JSON) somewhere that tests can subsequently access and Assert on. Note that since our tests use a fake in-memory server, the HttpClient used by WebHookNotificationSender will need to be reconfigured to use HostFixture.CreateClient so that the message can successfully be sent to the in-memory endpoint.

We will likely need to run these tests within the DisableParallelization collection so they run reliably.

### Guidance to review

As per changes.

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/Startup.cs
+++ b/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/Startup.cs
@@ -1,4 +1,5 @@
 using TeacherIdentity.AuthServer.Tests.Infrastructure;
+using TeacherIdentity.AuthServer.Tests.WebHooks;
 
 namespace TeacherIdentity.AuthServer.Tests;
 
@@ -10,11 +11,14 @@ public class Startup
         var dbHelper = new DbHelper(testConfiguration.Configuration.GetConnectionString("DefaultConnection") ??
             throw new Exception("Connection string DefaultConnection is missing."));
         var hostFixture = new HostFixture(testConfiguration, dbHelper);
-
         hostFixture.Initialize().GetAwaiter().GetResult();
+
+        var webhooksHostFixture = new WebHooksHostFixture(testConfiguration, dbHelper);
+        webhooksHostFixture.Initialize().GetAwaiter().GetResult();
 
         services.AddSingleton(testConfiguration);
         services.AddSingleton(dbHelper);
         services.AddSingleton(hostFixture);
+        services.AddSingleton(webhooksHostFixture);
     }
 }
--- a/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHookRequest.cs
+++ b/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHookRequest.cs
@@ -1,8 +1,8 @@
 namespace TeacherIdentity.AuthServer.Tests.WebHooks;
 
-public class WebHookRequest
+public record WebHookRequest
 {
-    public string? ContentType { get; set; }
-    public string? Signature { get; set; }
-    public string? Body { get; set; }
+    public required string ContentType { get; init; }
+    public required string Signature { get; init; }
+    public required string Body { get; init; }
 }
--- a/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHookRequest.cs
+++ b/dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHookRequest.cs
@@ -1,0 +1,8 @@
+namespace TeacherIdentity.AuthServer.Tests.WebHooks;
+
+public class WebHookRequest
+{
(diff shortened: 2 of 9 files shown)


hortha:
[Add e2e tests for webhooks](https://trello.com/c/hopv3ug2/939-add-e2e-tests-for-webhooks)

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/Startup.cs
```diff
@@ -10,11 +11,14 @@ public void ConfigureServices(IServiceCollection services)
         var dbHelper = new DbHelper(testConfiguration.Configuration.GetConnectionString("DefaultConnection") ??
             throw new Exception("Connection string DefaultConnection is missing."));
         var hostFixture = new HostFixture(testConfiguration, dbHelper);
-
         hostFixture.Initialize().GetAwaiter().GetResult();
 
+        var webhooksHostFixture = new WebHooksHostFixture(testConfiguration, dbHelper);
```
gunndabad, line 16: This is going to end up doing a lot of initialization work more than once. I wonder whether not inheriting from `HostFixture` for these, as you said, would be better.

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHookRequest.cs
```diff
@@ -0,0 +1,8 @@
+namespace TeacherIdentity.AuthServer.Tests.WebHooks;
+
+public class WebHookRequest
+{
+    public string? ContentType { get; set; }
+    public string? Signature { get; set; }
+    public string? Body { get; set; }
+}
```
gunndabad, line 8:
```suggestion
public record WebHookRequest
{
    public required string ContentType { get; init; }
    public required string Signature { get; init; }
    public required string Body { get; init; }
}
```

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHookRequestObserver.cs
```diff
@@ -0,0 +1,26 @@
+namespace TeacherIdentity.AuthServer.Tests.WebHooks;
+
+public class WebHookRequestObserver : IWebHookRequestObserver
+{
+    private readonly AsyncLocal<List<WebHookRequest>> _requests = new();
+
+    public void Initialize() => _requests.Value ??= new List<WebHookRequest>();
+
+    public Task OnWebHookRequestReceived(WebHookRequest webHookRequest)
```
gunndabad, line 9: Does this need to be `Task`-returning?

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHooksEndToEndTests.cs
```diff
@@ -0,0 +1,440 @@
...
+                    JsonNode.Parse(r.Body));
+            });
+        }
+        else
+        {
+            WebHookRequestObserver.AssertWebHookRequestsReceived(Array.Empty<Action<WebHookRequest>>());
+        }
+    }
+
+    private static string GenerateExpectedSignature(string secret, string content)
```
gunndabad, line 434: Could make the `CalculateSignature` method on `WebHookNotificationSender` `public static` and use that instead of duplicating here.

gunndabad (approved):

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHooksEndToEndTests.cs
hortha: Done

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/WebHooks/WebHookRequestObserver.cs
hortha: Nope

dotnet-authserver/tests/TeacherIdentity.AuthServer.Tests/Startup.cs
hortha: Done

PR #660, opened 2023-07-14
Improve TRN assignment support UI

gunndabad:
Until now we've had two separate flows; one for when a TRN was found and another for when it wasn't found. The latter wasn't actually linked to from the user detail page, only the user list page, so it wasn't easy to find. In addition, having both actions on each table row meant the text was overly-wrapped and didn't look great. The wording 'Mark as non DQT' wasn't amazing either.

This change consolidates both flows into a single one and adds a 'Does the user have a TRN?' question at the start. It's linked to from both the user list page and the detail page, labelled 'Assign TRN'.

I've also tweaked the information shown on the page after a TRN is entered:
- the NINO 'provided'/'not provided' row is gone (it's useless);
- ID DOB is added;
- DQT email is added;
- a warning is shown if the DQT email differs to the ID one (as it will be overwritten at the point the TRN is assigned).


gunndabad:
[Improve TRN assignment in Support UI](https://trello.com/c/NH61kpQ9/1114-improve-trn-assignment-in-support-ui)

dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Admin/AssignTrn/Index.cshtml.cs
```diff
@@ -42,23 +53,26 @@ public async Task<IActionResult> OnPost()
...
+            {
+                ModelState.AddModelError(nameof(Trn), "TRN does not exist");
+                return this.PageWithErrors();
+            }
 
-        if (otherOtherWithTrn is not null)
-        {
-            ModelState.AddModelError(nameof(Trn), "TRN is assigned to another user");
-            return this.PageWithErrors();
+            var otherOtherWithTrn = await _dbContext.Users.SingleOrDefaultAsync(u => u.Trn == Trn);
```
hortha, line 66: otherOther ? is this a multiverse related thing ? :)

hortha (approved):

PR #665, opened 2023-07-19
Backfill preferred name job

hortha:
### Context

We’ve added preferred name to the Core registration journey but some accounts pre-date that change (or have used the Legacy TRN journey which doesn’t capture preferred name) so don’t have a preferred name set. Before we overwrite names with DQT names we need to populate preferred name for all users.

Create a Hangfire job that can be manually triggered that sets a preferred name for all non-Staff users. The preferred name should be FirstName + " " + LastName. A UserUpdatedEvent should be created as appropriate. Existing preferred names should not be overwritten.

### Guidance to review

N/A

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


hortha:
[Back-fill preferred name](https://trello.com/c/f997Fuv0/1119-back-fill-preferred-name)

PR #696, opened 2023-08-04
Add TrnVerificationLevel to User

ihsanahmed-tpx:
### Context

Using trn verification will enable us to use a strict matching policy for claims

- added new prop to user
- added migration


dotnet-authserver/src/TeacherIdentity.AuthServer/Journeys/UserHelper.cs
```diff
@@ -139,6 +140,7 @@ public async Task<User> CreateUserWithTrnToken(AuthenticationState authenticatio
             LastSignedIn = _clock.UtcNow,
             RegisteredWithClientId = authenticationState.OAuthState?.ClientId,
             TrnLookupStatus = authenticationState.TrnLookupStatus,
+            TrnVerificationLevel = TrnVerificationLevel.Low
```
gunndabad, line 143: We should leave this as `null` for `CreateUserWithTrnToken`

dotnet-authserver/src/TeacherIdentity.AuthServer/Models/User.cs
```diff
@@ -36,4 +36,6 @@ public class User
     public string? MobileNumber { get; set; }
     public MobileNumber? NormalizedMobileNumber { get; set; }
     public bool TrnLookupSupportTicketCreated { get; set; }
+
```
gunndabad, line 39: Nit: don't need the newline here

ihsanahmed-tpx: lint fails if the newline is removed. Running `dotnet format` adds the newline in

gunndabad (approved):

PR #701, opened 2023-08-08
Option to update preferred name when changing official name

hortha:
### Context

We want to try to keep the preferred name in sync with official names. If users update their official name, we will prompt to see if they’d like to also update their preferred name.

Add a new page at /account/official-name/preferred-name following the designs at [Collecting a preferred name in addition to official name](https://tra-digital-design-history.herokuapp.com/get-an-identity/adding-preferred-name/#when-users-update-their-official-name) We should offer an option for the same combinations of first/middle/last names we do on the Change preferred name page. Redirect to this new page after /account/official-name/evidence.

If no option is chosen, show an error ‘Select which name to use’. If ‘Use a different preferred name’ is chosen, show a text box labelled ‘Your preferred name’. If this is empty, show an error ‘Enter your preferred name’.

If the name chosen is valid, redirect to /account/official-name/confirm and pass along the chosen preferred name. Update this page to update the user’s preferred name as well as submitting the official name change request to DQT.

### Guidance to review

N/A

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/IdentityLinkGenerator.cs
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/IdentityLinkGenerator.cs
@@ -205,22 +205,25 @@ public abstract class IdentityLinkGenerator
         Page("/Account/OfficialName/Index", authenticationJourneyRequired: false)
             .SetQueryParam(ClientRedirectInfo.QueryParameterName, clientRedirectInfo);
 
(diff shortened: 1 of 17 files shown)


hortha:
[Prompt to change preferred name when changing official name](https://trello.com/c/ABvD2kwQ/1145-prompt-to-change-preferred-name-when-changing-official-name)

gunndabad (approved):

PR #703, opened 2023-08-09
Prevent user changing their identity name when linked to a dqt record

hortha:
### Context

When we have DQT name synchronisation turned on we should not allow user’s to change their ID name directly.

Amend the /account/name and /account/name/confirm pages to check if the user has a TRN linked and return a 400 response if they do if the DQT name synchronisation feature is enabled.

A filter attribute would work well here (see CheckOfficialDateOfBirthChangeIsEnabledAttribute for some prior art).

### Guidance to review

Simplish change. Does not include code to stop links being clicked in the account page as this will be part of another card.

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


hortha:
[Prevent user changing their ID name when linked to a DQT record](https://trello.com/c/BKOLIsnA/1150-prevent-user-changing-their-id-name-when-linked-to-a-dqt-record)

PR #705, opened 2023-08-10
Amended account landing page to reflect most recent design history

hortha:
### Context

We have a revised version of the account page that better fits when we’ve got DQT name synchronisation enabled.

Amend the account page so that when the DQT name synchronisation feature is enabled we show a revised version of the page as shown on [Avoiding duplicate names](https://tra-digital-design-history.herokuapp.com/get-an-identity/avoiding-duplicate-names/)

The page header should be ‘Confirm your details’.

Add the ‘Check your details we have for you from your DfE teaching record’ and ‘You’ll need to provide evidence for some name changes and all date of birth changes.' paragraphs.

Add the ‘Teaching record’ block with first, middle, last names and date of birth. The Change links for the name rows should go to /account/official-name if the user has a TRN linked, otherwise to /account/name. If the DQT DOB and ID DOB's don't match then show both in their own rows, with the appropriate Change link for each. If the DOB's do match, do not show a Change link. If the DQT DOB is pending review show the 'PENDING REVIEW' tag.

Add the ‘Account details’ blow with rows for preferred name, email and phone number with the appropriate change links.

We should retain the ‘Confirm your correct date of birth’ notification banner if the DQT DOB and ID DOB’s do not match.

Unlike the design history post, we should not have a ‘Confirm’ button but keep the ‘Back to <client>’ button we have today.

### Guidance to review

N/A

### Checklist

-   [x] Attach to Trello card
-   [x] Rebased master

### DIFF
--- a/dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Account/Index.cshtml
+++ b/dotnet-authserver/src/TeacherIdentity.AuthServer/Pages/Account/Index.cshtml
@@ -1,7 +1,7 @@
 @page "/account"
 @model TeacherIdentity.AuthServer.Pages.Account.IndexModel
 @{
-    ViewBag.Title = "DfE Identity account";
+    ViewBag.Title = "Confirm your details";
 
     var dateOfBirthChangeEnabled = Model.DqtDateOfBirth is null || (Model.DateOfBirthConflict && !Model.PendingDqtDateOfBirthChange);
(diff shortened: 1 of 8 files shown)


hortha:
[Revised account page for ID name changes](https://trello.com/c/K6gOI9Xz/1149-revised-account-page-for-id-name-changes)

PR #712, opened 2023-08-15
Added extra client side rate limiting to reduce the load on DQT API

hortha:
### Context

While running the user import on a large file from NPQ, eventually the user import process hit the rate limiting on the DQT API. This would also mean that other calls to DQT API from other services might be affected which is bad.

Add internal throttling within the user import process so that it will never get near the rate limit when calling DQT API (e.g. around 1/2 the value).

Also amend the user import to short circuit ignoring rows of data which have already been processed, if a job gets retried by Hangfire.

### Guidance to review

Change to just the user import processor

### Checklist

-   [x] Attach to Trello card
-   [ ] Rebased master


hortha:
[Add throttling for DQT API calls in User Import](https://trello.com/c/dAxCnJtF/1175-add-throttling-for-dqt-api-calls-in-user-import)

gunndabad (approved):

PR #719, opened 2023-09-11
Add content to landing page when we require TRN and NINO to match

gunndabad:
For Claim and AYTQ we require a match on TRN and NINO. This adds content to the landing page to tell users that they will need this information.

We have two landing pages that are very similar; I've resisted the urge to unify them here


gunndabad:
[Add additional content to landing page for Strict TRN lookup requirement](https://trello.com/c/hMN48ZpP/1215-add-additional-content-to-landing-page-for-strict-trn-lookup-requirement)

MrKevJoy (approved):

PR #778, opened 2024-01-12
Fix elevating TrnVerificationLevel when matching an existing account

gunndabad:


gunndabad:
[Fix missing TRN in access token when registering user matches an existing account](https://trello.com/c/9H4wBA7i/1629-fix-missing-trn-in-access-token-when-registering-user-matches-an-existing-account)

hortha (approved):

PR #790, opened 2024-04-10
Update content for users who weren't automatically matched

gunndabad:


gunndabad:
[Content change for users who could not be verified](https://trello.com/c/pp18gIWk/1970-content-change-for-users-who-could-not-be-verified)

MrKevJoy (approved):

